PR #2, opened 2023-06-08
graphql-apis

knockknockyoo:


services/data-interface.js
AustinSMueller, line 1: We need to update the workflow related fields in the Application as it progresses through the workflow. A document with all of the state transitions can be found [here](https://nih.sharepoint.com/:x:/r/sites/NCI-CBIITCTOSProjectManagement/Shared%20Documents/CRDC%20DCC%20aka%20Data%20Hub/Software%20Development/BE/CRDC%20DataHub%20Pre-Submission%20Form%20State%20Transitions.xlsx?d=wdce87448ad8243ccb5fc13628a92ee41&csf=1&web=1&e=d7LFXE). 

The "status", "history", and "updatedAt" fields will need to be updated with each state transition but there may other query specific values that will need to be updated as well

AustinSMueller, line 1: Please verify that the Application is in the correct state before executing each state transition. If it is not then please return a descriptive error message

AustinSMueller (changes requested):
Please see the comments on the data interface file for a description of the requested changes

PR #7, opened 2023-06-23
Email apis

knockknockyoo:


config.js
AustinSMueller, line 1: Please update "env.template" and "readme.md" with the new environment variables

services/notifications.js
AustinSMueller, line 1: I know both notifications.js and notify.js were reused from Bento but it's unclear what each of these classes responsibilities it. Can we consolidate into one class or rename in a way that makes it clear what the responsibilities of each is?

AustinSMueller (changes requested):

env.template
AustinSMueller, line 1: EMAILS_ENABLED variable is missing from this file and the README

services/email.js
```diff
@@ -0,0 +1,54 @@
+const { createTransport } = require('nodemailer');
+const config = require('../config');
+
+class EmailService {
+
+    constructor() {
```
AustinSMueller, line 6: When instantiating a service, we should try not to import config and instead pass the config variables into the constructor so that we can make multiple different versions if needed. Can you remove the config import and add email transport to the constructor?

knockknockyoo: I've updated the changes.

PR #10, opened 2023-07-03
Inactive Applications

knockknockyoo:


app.js
```diff
@@ -36,6 +42,18 @@ app.use(createSession(config.session_secret, config.session_timeout, config.mong
 // add graphql endpoint
 app.use("/api/graphql", graphqlRouter);
 
+// Scheduled cronjob once a day (1am) eastern time
+cronJob.schedule("1 0 1 * * *", async () => {
```
AustinSMueller, line 46: Based on feedback from when we implemented scheduled jobs in Bento, we should make the cronJob schedule parameter an environment variable so that we can setup test environments to run more frequently without having a code change

config.js
```diff
@@ -4,7 +4,7 @@ let config = {
     //info variables
     version: process.env.VERSION,
     date: process.env.DATE,
-
+    inactive_user_days : process.env.INACTIVE_APPLICATION_DAYS ? process.env.INACTIVE_APPLICATION_DAYS : 45,
```
AustinSMueller, line 7: This pattern of assigning a value to a variable if it is null can be replaced with:
`inactive_user_days : process.env.INACTIVE_APPLICATION_DAYS || 45`

services/application.js
```diff
@@ -143,6 +143,27 @@ class Application {
...
+
+    async deleteInactiveApplications(inactiveDays) {
+        const inactiveCondition = {
+            updatedAt: {
+                $lt: subtractDaysFromNow(inactiveDays)
+            },
+            status: SUBMITTED
+        };
+        const applications = await this.applicationCollection.aggregate([{$match: inactiveCondition}, {"$limit": 1}]);
+        if (applications.length > 0) {
```
AustinSMueller, line 155: let's make this check null safe and throw an error message if the search fails and "applications" is null:
`if (!applications) throw new Error("error message here")`
`if (applications?.length > 0) {`
`...`

utility/time-utility.js
AustinSMueller, line 1: Please add some unit tests for this new function

services/application.js
AustinSMueller, line 1: Please add some unit tests for deleteInactiveApplications

AustinSMueller (changes requested):

services/application.js
knockknockyoo: Added

utility/time-utility.js
knockknockyoo: Added

services/application.js
knockknockyoo: Updated.

config.js
knockknockyoo: Updated.

app.js
knockknockyoo: Updated.

PR #11, opened 2023-07-05
Submit email notification

knockknockyoo:


services/application.js
```diff
@@ -78,6 +82,7 @@ class Application {
     }
 
     async submitApplication(params, context) {
+        await this.userService.getMyUser(params, context);
```
AustinSMueller, line 85: I think we can remove this getMyUser call. We are expecting that getMyUser will be called at login so it should already be in the session. If it is not then it will fail the verification in the lines below

services/application.js
```diff
@@ -135,7 +141,7 @@ class Application {
             $set: {reviewComment: document.comment, wholeProgram: document.wholeProgram, status: APPROVED, updatedAt: history.dateTime},
             $push: {history}
         });
-        return (updated.modifiedCount && updated.modifiedCount > 0) ? await this.dbService.find(APPLICATION, {_id: document._id}) : null;
+        return updated.modifiedCount && updated.modifiedCount > 0 ? this.getApplicationById(document._id) : null;
```
AustinSMueller, line 144: Can we make this null safe by changing to "updated?.modifiedCount"? "updated" should never be null but we should use null safe checks wherever possible

services/application.js
```diff
@@ -149,7 +155,7 @@ class Application {
             $set: {reviewComment: document.comment, status: REJECTED, updatedAt: history.dateTime},
             $push: {history}
         });
-        return (updated.modifiedCount && updated.modifiedCount > 0) ? await this.dbService.find(APPLICATION, {_id: document._id}) : null;
+        return updated.modifiedCount && updated.modifiedCount > 0 ? this.getApplicationById(document._id) : null;
```
AustinSMueller, line 158: Let's make this null safe as well, the same as the previous comment

AustinSMueller (changes requested):
The submodule reference has not been updated. It is currently pointing to this [commit](https://github.com/CBIIT/crdc-datahub-database-drivers/commit/f7efcae4e7af2c3fbbd85bc21c45ed28d72ef2fd) which is missing the users service entirely

services/application.js
knockknockyoo: updated

services/application.js
knockknockyoo: updated

services/application.js
knockknockyoo: updated

PR #15, opened 2023-07-20
Crdcdh 43

Wesley-Lau:


services/application.js
```diff
@@ -22,6 +24,7 @@ class Application {
     }
 
     async getApplicationById(id) {
+        console.log('test')
```
AustinSMueller, line 27: Please remove this "test" output

config.js
```diff
@@ -18,6 +18,9 @@ let config = {
     email_transport: getTransportConfig(),
     emails_enabled: process.env.EMAILS_ENABLED ? process.env.EMAILS_ENABLED.toLowerCase() === 'true' : true,
     emails_url: process.env.EMAIL_URL ? process.env.EMAIL_URL : 'http://localhost:4010',
+    // temp url
+    temp_email_receiver:"lauwc@nih.gov",
+    temp_test_url: "https://datacommons.cancer.gov/",
```
AustinSMueller, line 23: These added variables should be configured as environment variables, given more descriptive names, and also added to the README.md file

I would recommend using SUBMISSION_DOC_URL to replace temp_test_url and should have a default value of an empty string

services/application.js
```diff
@@ -191,6 +197,20 @@ class Application {
...
+        await this.notificationService.approveQuestionNotification(application?.primaryContact?.email,
+            // Organization Owner and concierge assigned/Super Admin
+            config.temp_email_receiver,
+        {
+            firstName: application?.applicantName
+        }, {
+            study: application?.study?.name,
+            doc_url: config.temp_test_url,
+            org_owner_email: config.temp_email_receiver,
+            concierge_email: config.temp_email_receiver
```
AustinSMueller, line 211: Getting org owner and concierge email will require new user operations. I will create a task to implement this that will be need to be completed before we can finish this change.

AustinSMueller: The below logic should be added to determine who should be in the CC list

if the user does not have an org or the org does not have a concierge then CC the admin else CC the concierge
if the user has an org and the org has an owner then CC the org owner
Pass empty strings as the message parameters for org owner or concierge if they are missing

AustinSMueller (changes requested):

PR #28, opened 2023-08-08
added to filter valid application status

knockknockyoo:
Deleted Application is not allowed to be listed.


n2iw (changes requested):
Please add **New** status in the list, thanks!

n2iw (approved):
LGTM

PR #44, opened 2023-08-31
Added email notification for approved email.

Wesley-Lau:


services/application.js
AustinSMueller, line 1: I think there were a lot of schema changes since this was originally written. This will need to be re-written using the steps below:

- Get the submitter's email from application.applicant.applicantEmail and add it to the list that will be used for the notification's "TO" field
- Get the submitter's ID from application.applicant.applicantID
- use the submitter's ID to lookup the submitter's user object
- if the submitter's user object has an organization, then store the org's ID
- use the org ID to lookup the organization object
- if the organization has an owner, then use the ID in organization.owner to lookup the owner's user object and store that in the list for notification "CC"
- if the organization has an curator, then use the ID in organization.concierges to lookup the curator's user object and store that in the list for notification "CC"
- if the application does not have an organization or the organization does not have a curator then add all admins to the list for notification "CC"
- send the email using the "TO" and "CC" lists

AustinSMueller (changes requested):

services/application.js
```diff
@@ -346,7 +347,58 @@ class Application {
...
+
 
+        await this.notificationService.approveQuestionNotification(application?.applicant?.applicantEmail,
+            // Organization Owner and concierge assigned/Super Admin
+            `${org_owner_email} ; ${cc_email}`,
+        {
+            firstName: application?.applicant?.applicantName
+        }, {
+            study: application?.studyAbbreviation,
+            doc_url: config.submission_doc_url,
```
AustinSMueller, line 397: "config" is undefined here and needs to be imported

AustinSMueller, line 397: please add submission_doc_url to config.js, the env.template, and the readme doc

AustinSMueller (changes requested):

PR #46, opened 2023-08-31
Update Timestamp to ISO9601

knockknockyoo:


app.js
```diff
@@ -14,7 +14,7 @@ const {DATABASE_NAME, APPLICATION_COLLECTION, USER_COLLECTION, ORGANIZATION_COLL
 const {Application} = require("./services/application");
 const {MongoQueries} = require("./crdc-datahub-database-drivers/mongo-queries");
 const {DatabaseConnector} = require("./crdc-datahub-database-drivers/database-connector");
-const {getCurrentTimeYYYYMMDDSS} = require("./utility/time-utility");
+const {getCurrentTimeYYYYMMDDSS} = require("./crdc-datahub-database-drivers/utility/time-utility");
```
n2iw, line 17: Function needs a better name, see PR comment in submodule.

services/organization.js
```diff
@@ -1,14 +1,21 @@
...
     }
 
     async getOrganizationByID(id) {
         let result = await this.organizationCollection.find(id);
-        return result?.length > 0 ? result[0] : null;
+        return result?.length > 0 ? toISOTime(result[0]) : null;
     }
 }
 
+const toISOTime = (aOrg) => {
```
n2iw, line 13: This function name is a bit misleading, people may think we are converting the object into an ISO time. I suggest rename it to something like transformDateTime, transformDateTimeFormat or similar.

services/application.js
```diff
@@ -363,6 +360,18 @@ function verifyReviewerPermission(context){
         .verifyRole([ROLES.ADMIN, ROLES.FEDERAL_LEAD]);
 }
 
+const toISOTime = (aApp) => {
```
n2iw, line 363: This function name is a bit misleading, people may think we are converting the object into an ISO time. I suggest rename it to something like transformDateTime, transformDateTimeFormat or similar.

services/application.js
```diff
@@ -209,7 +206,7 @@ class Application {
         const deletedOne = await this.getApplicationById(document._id);
         let result = null;
         if (deletedOne && await this.dbService.deleteOne(APPLICATION, {_id: document._id})) {
-            result = deletedOne[0];
```
n2iw, line 212: Is this change from **deleteOne[0]** to **deleteOne** intentional?

Also, we should NOT delete any application records from DB, should only change status to "Deleted" instead.

n2iw (changes requested):
Please see individual comments.

app.js
knockknockyoo: updated!

services/organization.js
knockknockyoo: updated!

services/application.js
knockknockyoo: updated!

services/application.js
knockknockyoo: It was on our to-do list. I have now pushed it for an update.

crdc-datahub-database-drivers
n2iw, line 1: PR in submodule has been merged to 1.0.0 branch, please update submodule to the latest 1.0.0, thanks!

services/application.js
n2iw: deleteApplication function needs permission control, please add a TODO comment in the function, so we won't forget about it.

n2iw (changes requested):
Please see two new comments, thanks!

knockknockyoo:
> 

Thank you. I've added a TODO comment.

PR #49, opened 2023-09-05
Send Email Before Application Expired

knockknockyoo:


services/application.js
```diff
@@ -303,27 +303,28 @@ class Application {
         }
     }
 
+    async remindApplicationSubmission() {
+        const inactiveDuration = this.emailParams.inactiveDays - this.emailParams.remindDay;
```
AustinSMueller, line 307: This logic is incorrect. If use default values, inactiveDays is 45 and remnidDay is 30 so inactiveDuration is 15 and we send a reminder after 15 days. We want to send the reminder after 30 days so we can remove this "inactiveDuration" calculation and just use the remindDay value instead

services/application.js
```diff
@@ -303,27 +303,28 @@ class Application {
...
+        const remindCondition = {
+            updatedAt: {
+                $lt: subtractDaysFromNow(inactiveDuration),
```
AustinSMueller, line 310: This was not working for me, the query was always returning 0 applications. I think the issue is that subtractDaysFromNow is returning a date object. I was able to get it to work by converting to an ISO string before using it in the pipeline here.  ".toISOString()"

AustinSMueller (changes requested):

app.js
```diff
@@ -55,12 +55,14 @@ cronJob.schedule(config.schedule_job, async () => {
...
         const organizationCollection = new MongoDBCollection(dbConnector.client, DATABASE_NAME, ORGANIZATION_COLLECTION);
-        const emailParams = {url: config.emails_url, officialEmail: config.official_email, inactiveDays: config.inactive_user_days};
+        const emailParams = {url: config.emails_url, officialEmail: config.official_email, inactiveDays: config.inactive_user_days, remindDay: config.remind_application_days};
         const logCollection = new MongoDBCollection(dbConnector.client, DATABASE_NAME, LOG_COLLECTION);
         const userService = new User(userCollection, logCollection);
         const dataInterface = new Application(logCollection, applicationCollection, new Organization(organizationCollection), userService, dbService, notificationsService, emailParams);
         console.log("Running a scheduled background task to delete inactive application at " + getCurrentTime());
         await dataInterface.deleteInactiveApplications(config.inactive_user_days);
+        console.log("Running a scheduled background task to remind inactive application at " + getCurrentTimeYYYYMMDDSS());
```
AustinSMueller, line 64: This  getCurrentTimeYYYYMMDDSS() needs to be updated to getCurrentTime()

services/application.js
knockknockyoo: updated it to 30days.

services/application.js
knockknockyoo: updated it to 30 days.

AustinSMueller (approved):

PR #50, opened 2023-09-05
Crdcdh 324

AustinSMueller:


services/application.js
```diff
@@ -26,7 +26,13 @@ class Application {
     async getApplication(params, context) {
         verifySession(context)
             .verifyInitialized();
-        return await this.getApplicationById(params._id);
+        const application = await this.getApplicationById(params._id);
+        const isAdminOrFedLead = [USER.ROLES.ADMIN, USER.ROLES.FEDERAL_LEAD].includes(context.userInfo?.role);
+        const isSubmitter = application?.applicant?.applicantID === context?.userInfo?._id;
+        if (!isAdminOrFedLead && !isSubmitter){
```
n2iw, line 32: I think it would be better to revert the logic, so that only submitter, admin or fed lead can see an application, everything else is blocked by default, this way will be a little more secure.

n2iw (dismissed):
Please check if this PR will fix [CRDCDH-323](https://tracker.nci.nih.gov/browse/CRDCDH-323). I suspect it will at least partially fix it.

PR #52, opened 2023-09-07
Removed ISO Conversion

knockknockyoo:
- removed ISO conversion
- used scalar Date
- date automatically converted to string


n2iw (approved):
LGTM

PR #54, opened 2023-09-08
Fixed saveApplication issues

AustinSMueller:
- save application was not returning the application when updating an application
- a state change update log event was being created for each API call even when the state was not changed


utility/application-util.js
```diff
@@ -0,0 +1,31 @@
+const {IN_PROGRESS} = require("../constants/application-constants");
```
knockknockyoo, line 1: Let's refrain from developing an application utility as it disrupts the integrity of the Application capsule. We should include outside of application.js

services/application.js
```diff
@@ -97,25 +98,20 @@ class Application {
...
+        const id = inputApplication?._id;
+        if (!id) {
+            return await this.createApplication(inputApplication, context.userInfo);
+        }
+        const storedApplication = await this.getApplicationById(id);
+        const prevStatus = storedApplication?.status;
+        let application = {...storedApplication, ...inputApplication, status: IN_PROGRESS};
+        application = await updateApplication(this.applicationCollection, application, prevStatus, context?.userInfo?._id);
+        if (prevStatus !== application.status){
+            await logStateChange(this.logCollection, context.userInfo, application, prevStatus);
```
knockknockyoo, line 112: I'm okay with using all asynchronous methods, but running four promises in a row might slow things down.

AustinSMueller: In this instance it is going to run the 1 createApplication async method or the other 3 async methods. I think these need to be made synchronous with await because each relies on the success and return values or return values the previous call. If getApplicationByID does not find an application then an error should be thrown and we should not do the update operation. If the updateApplication fails then we do not want to log that an update took place

utility/application-util.js
AustinSMueller: The tradeoff with encapsulation is that we lose our ability to unit test these functions. In general we do not want these to be called outside of the application class but they need to be accessible to the unit testing file. I can move these within the application class but then we run into the issue of running static methods within an instantiated class. I am fine with removing this file but I don't think it will solve the issue. I don't know if there is a way keep these scoped to the application capsule while maintaining the unit testing capability

utility/application-util.js
```diff
@@ -0,0 +1,31 @@
...
+const {HistoryEventBuilder} = require("../domain/history-event");
+const ERROR = require("../constants/error-constants");
+const {UpdateApplicationStateEvent} = require("../crdc-datahub-database-drivers/domain/log-events");
+
+
+async function updateApplication(applicationCollection, application, prevStatus, userID) {
+    if (prevStatus !== IN_PROGRESS) {
```
knockknockyoo, line 8: If we share this function, we should include parameter instead of 'IN_PROGRESS'.

utility/application-util.js
AustinSMueller: I've moved these functions to application.js to preserve encapsulation for now and removed the unit tests. The unit tests in this case were very minor but we will need to either find a way to unit test private methods in the future or expose them to the test files

PR #60, opened 2023-09-13
Presigned url

knockknockyoo:
Added test-codes
Added aws config environments


package.json
```diff
@@ -8,6 +8,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "aws-sdk": "^2.1457.0",
```
n2iw, line 11: Is this dependency needed? Since we are doing S3 related operations in the submodule.

n2iw (changes requested):
- Please take a look at the comment for the dependency
- Please update the submodule to latest 1.0.0 branch, since I've merged the PR there.

package.json
knockknockyoo: This library requires to run the S3 operation in the submodule.

knockknockyoo:
I've updated the submodule.

n2iw (approved):
LGTM

PR #69, opened 2023-09-18
Store Approved Studies

knockknockyoo:
1. Created Store Approved Service
2. JSON Pase test codes
3. Approved Service Error code


knockknockyoo:
It is currently verifying that the study name or its abbreviation are unique to prevent the storage of duplicate data.

PR #74, opened 2023-09-21
Create batch API

knockknockyoo:
1. Added Batch Domain
3. Added a verify batch function before db insertion & batch verify error constant
4. Added Batch service test codes


vshand11:
Hi Young,  updates required 1) In batch-service.js, needs to insert files, file count for both metadata and file types.  2) in graphql-router.js, needs to clean up codes by move anonymous function of createBatch into application class for the existing pattern. 3) in graphql-router.js, move verifyBatchPermission into user-info-verifier.js for the existing pattern.

knockknockyoo:
I've applied the change. Please, take a moment for another review. Thanks.

vshand11:
Great! Much clean now.   One enhancement in the function, createPrefix in the batch-servive.js file.   Please check if the submission doc found by submissionID has rootPath that the real path FE used for uploading files.   if rootPath is not null then use rootPath, otherwise use default orgID/sussionID.

vshand11:
One property is set right when creating the Batch, fileCount: Int # calculated by BE.  
Right now it's 0 always.

vshand11:
The verifyBatchPermission function needs to be optimized.   If one of org owners matches the login user, then allowed to create batch.   The current logic only check the first org owner, that only works if there is only one owner.

vshand11:
Passed my review and testing.  Codes are merged to 2.0.0

PR #82, opened 2023-09-26
Update org owner

Wesley-Lau:


services/application.js
```diff
@@ -339,16 +339,14 @@ class Application {
...
         // org owner email
-        let org = await this.organizationService.getOrganizationByID(application?.organization?._id);
         let org_owner_email = null
-        let org_owner_id = org?.owner
-        if(org_owner_id){
-            let org_owner = await this.userService.getUserByID(org_owner_id);
-            if(org_owner?.email){
-                org_owner_email = org_owner?.email
-            }
+        let orgOwner = await this.userService.getOrgOwner(application?.organization?._id)
```
AustinSMueller, line 343: This code results in a list of org owners that looks like this: `nullorgowner@email.com ; orgowner2@email.com` this can be replaced by using `orgOwner.join(";")`

services/application.js
```diff
@@ -370,8 +368,6 @@ class Application {
         }else{
             cc_email = admin_email
         }
-        // submission documentation
```
AustinSMueller, line 373: after removing this line, the code fails on line 374 because sub_doc_url is undefined

services/application.js
```diff
@@ -339,16 +339,14 @@ class Application {
...
-            }
+        let orgOwner = await this.userService.getOrgOwner(application?.organization?._id)
+        for(let i of orgOwner){
+            org_owner_email = org_owner_email  + i.email + " ; "
         }
+        
         // concierge email
+        let org = await this.organizationService.getOrganizationByID(application?.organization?._id);
         let concierge_email = null
         let org_concierge_id = org?.concierges
```
AustinSMueller, line 351: concierge (data curators) should be looked up using a query very similar to the org owner lookup you implemented in this ticket. They can be searched by looking up 
`"$match": {
                "organization.orgID": orgID,
                role: "Data Curator"
            }`
 please test this to make sure it works after implementing

AustinSMueller (changes requested):

PR #84, opened 2023-09-27
Crdcdh list approved studies

mtangmt:
This adds the listApprovedStudies endpoint as outlined in ticket 114, and it just returns all approved studies in the collection.


n2iw:
There is another PR for the same functionality, we will use this other one and close this PR.

PR #89, opened 2023-10-04
Crdcdh 310 - `listSubmissions`

mtangmt:
This PR creates a listSubmissions api endpoint, and a temporary createDataSubmission api endpoint.
This should fulfill listSubmissions requirements.
This does not fulfill all the requirements for createSubmission.


services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let dataSubmission=params.dataSubmission;
+        let userInfo = context.userInfo;
+        let newApplicationProperties = {
+            _id: v4(undefined, undefined, undefined),
+            name: params.name,
+            submitterID: userInfo._id,
+            submitterName: formatApplicantName(userInfo),
+            organization: {_id:"testid", name: userInfo.organization.orgName},
+            dataCommons: "CDS",
```
knockknockyoo, line 36: Please, DO NOT hard-coded this constant.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    async createDataSubmission(params, context) {
+        verifySession(context)
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let dataSubmission=params.dataSubmission;
+        let userInfo = context.userInfo;
+        let newApplicationProperties = {
+            _id: v4(undefined, undefined, undefined),
```
knockknockyoo, line 31: Could you verify the difference v4 vs v4(undefined, undefined, undefined)?

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            organization: {_id:"testid", name: userInfo.organization.orgName},
+            dataCommons: "CDS",
+            modelVersion: "string for future use",
+            studyAbbreviation: params.studyAbbreviation,
+            dbGapID: params.dbGapID,
+            bucketName: "todo: get from database",
+            rootPath: "todo: organization/study?",
+            status: NEW,
+            history: [HistoryEventBuilder.createEvent(userInfo._id, NEW, null)],
+            concierge: "todo: get from database?",
```
knockknockyoo, line 44: Could you please use a "// TODO" comment instead? The IDE will identify and track these TODO tasks.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+        this.notificationService = notificationsService;
+        this.emailParams = emailParams;
+    }
+
+    async createDataSubmission(params, context) {
+        verifySession(context)
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let dataSubmission=params.dataSubmission;
+        let userInfo = context.userInfo;
```
knockknockyoo, line 29: This should be const. I believe.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            createdAt: getCurrentTime(),
+            updatedAt: getCurrentTime()
+        };
+
+        dataSubmission = {
+            ...dataSubmission,
+            ...newApplicationProperties
+        };
+        const res = await this.dataSubmissionCollection.insert(dataSubmission);
+        // if (res?.acknowledged) await this.logCollection.insert(CreateApplicationEvent.create(userInfo._id, userInfo.email, userInfo.IDP, dataSubmission._id));
```
knockknockyoo, line 54: Could you remove unnecessary comment? Or, leave a TODO comment for future use.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+
+
+    listConditions(userID, userRole, aUserOrganization, params) {
+        const validApplicationStatus = {status: {$in: [NEW, IN_PROGRESS, SUBMITTED, RELEASED, COMPLETED, ARCHIVED]}};
+        // Default conditons are:
+        // Make sure application has valid status
+        let conditions = [{$and: [validApplicationStatus]}];
+        
+        // Filter on organization and status
+        if (params.organization !== "All") {
```
knockknockyoo, line 66: Please, DO NOT use hard-coded constant.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            conditions[0].$and.push({"organization.name": params.organization});
+        }
+        if (params.status !== "All") {
```
knockknockyoo, line 69: Please, DO NOT use hard-coded constant.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            conditions[0].$and.push({status: params.status});
+        }
+        // List all applications if Fed Lead / Admin / Data Concierge / Data Curator
+        const listAllApplicationRoles = [USER.ROLES.ADMIN, USER.ROLES.FEDERAL_LEAD, USER.ROLES.CURATOR, USER.ROLES.DC_POC];
+        if (listAllApplicationRoles.includes(userRole)) {
+            return [{"$match": {"$or": conditions}}];;
```
knockknockyoo, line 75: Could you remove ;?

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+        }
+        // If org owner, add condition to return all data submissions associated with their organization
+        if (userRole === USER.ROLES.ORG_OWNER && aUserOrganization?.orgID) {
+            conditions[0].$and.push({"organization.name": aUserOrganization.orgName});
+            return [{"$match": {"$or": conditions}}];;
```
knockknockyoo, line 80: Could you remove ;?

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            ...dataSubmission,
+            ...newApplicationProperties
+        };
+        const res = await this.dataSubmissionCollection.insert(dataSubmission);
+        // if (res?.acknowledged) await this.logCollection.insert(CreateApplicationEvent.create(userInfo._id, userInfo.email, userInfo.IDP, dataSubmission._id));
+        return dataSubmission;
+    }
+
+
+    listConditions(userID, userRole, aUserOrganization, params) {
```
knockknockyoo, line 59: Could you declare this function outside of this class?

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+class DataSubmission {
+    constructor(logCollection, dataSubmissionCollection, organizationService, userService, dbService, notificationsService, emailParams) {
+        this.logCollection = logCollection;
+        this.dataSubmissionCollection = dataSubmissionCollection;
+        this.organizationService = organizationService;
+        this.userService = userService;
+        this.dbService = dbService;
+        this.notificationService = notificationsService;
+        this.emailParams = emailParams;
+    }
```
knockknockyoo, line 22: Please, remove unused fields.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+const {v4} = require('uuid')
+const {getCurrentTime, subtractDaysFromNow} = require("../crdc-datahub-database-drivers/utility/time-utility");
+const {HistoryEventBuilder} = require("../domain/history-event");
+const {verifySession} = require("../verifier/user-info-verifier");
+const ERROR = require("../constants/error-constants");
+const {getSortDirection} = require("../crdc-datahub-database-drivers/utility/mongodb-utility");
+const USER_CONSTANTS = require("../crdc-datahub-database-drivers/constants/user-constants");
+const {USER} = require("../crdc-datahub-database-drivers/constants/user-constants");
+const ROLES = USER_CONSTANTS.USER.ROLES;
+
```
knockknockyoo, line 12: Please, delete unused imports.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let dataSubmission=params.dataSubmission;
+        let userInfo = context.userInfo;
+        let newApplicationProperties = {
+            _id: v4(undefined, undefined, undefined),
+            name: params.name,
+            submitterID: userInfo._id,
+            submitterName: formatApplicantName(userInfo),
+            organization: {_id:"testid", name: userInfo.organization.orgName},
```
knockknockyoo, line 35: If a user belongs to any organization, it needs to look up the database by the orgName if possible.

services/data-submission.js
```diff
@@ -0,0 +1,131 @@
...
+
+}
+
+function formatApplicantName(userInfo){
+    if (!userInfo) return "";
+    let firstName = userInfo?.firstName || "";
+    let lastName = userInfo?.lastName || "";
+    lastName = lastName.trim();
+    return firstName + (lastName.length > 0 ? " "+lastName : "");
+}
```
knockknockyoo, line 126: This is duplicates codes in application.js. It should create a public function if possible.

knockknockyoo (commented):
It looks good. I've left some comments below.

services/data-submission.js
mtangmt: User info can be const. dataSubmission is reassigned later, so it must be a let.

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+            modelVersion: "string for future use",
+            studyAbbreviation: params.studyAbbreviation,
+            dbGaPID: params.dbGaPID,
+            // TODO: get bucket name, and rootPath from organization database
+            bucketName: "get from database",
+            rootPath: "organization/study?",
+            status: NEW,
+            history: [HistoryEventBuilder.createEvent(userInfo._id, NEW, null)],
+            // TODO: get conceirge data from organization database
+            concierge: "get from organization database?",
```
knockknockyoo, line 76: can you write a TODO comment too?

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+            name: params.name,
+            submitterID: userInfo._id,
+            submitterName: formatApplicantName(userInfo),
+            organization: {_id: userInfo.organization.orgID, name: userInfo.organization.orgName},
+            dataCommons: "CDS",
+            modelVersion: "string for future use",
+            studyAbbreviation: params.studyAbbreviation,
+            dbGaPID: params.dbGaPID,
+            // TODO: get bucket name, and rootPath from organization database
+            bucketName: "get from database",
```
knockknockyoo, line 71: can you write a TODO comment too?

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+        this.organizationService = organizationService;
+        this.userService = userService;
+    }
+
+    async createSubmission(params, context) {
+        // TODO: Add this requirement: Study abbreviation must have an approved application within user's organization
+        verifySession(context)
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let submission = params.submission;
```
knockknockyoo, line 58: I think this is not necessary. Could you double check?

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+        const userInfo = context.userInfo;
+        let newApplicationProperties = {
+            _id: v4(),
+            name: params.name,
+            submitterID: userInfo._id,
+            submitterName: formatApplicantName(userInfo),
+            organization: {_id: userInfo.organization.orgID, name: userInfo.organization.orgName},
```
knockknockyoo, line 65: On my local machine, organization.orgID is always null. Could you double check?

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+
+
+function listConditions(userID, userRole, aUserOrganization, params){
+    const validApplicationStatus = {status: {$in: [NEW, IN_PROGRESS, SUBMITTED, RELEASED, COMPLETED, ARCHIVED]}};
+    // Default conditons are:
+    // Make sure application has valid status
+    let conditions = [{$and: [validApplicationStatus]}];
+    
+    // Filter on organization and status
+    if (params.organization !== ALL_FILTER) {
```
knockknockyoo, line 20: If organization is null or undefined, this will cause unexpected errors.

knockknockyoo (changes requested):

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+const ROLES = USER_CONSTANTS.USER.ROLES;
+const ALL_FILTER = "All";
+
+
+
+function listConditions(userID, userRole, aUserOrganization, params){
+    const validApplicationStatus = {status: {$in: [NEW, IN_PROGRESS, SUBMITTED, RELEASED, COMPLETED, ARCHIVED]}};
+    // Default conditons are:
+    // Make sure application has valid status
+    let conditions = [{$and: [validApplicationStatus]}];
```
knockknockyoo, line 17: Could you think about using an object instead? It looks fine, but it might not be very intuitive at first. I think you no longer need an $or query after making adjustments
let condition = { ...validApplicationStatus };
const listAllApplicationRoles = ["ADMIN", "FEDERAL_LEAD", "CURATOR", "DC_POC"];
if (listAllApplicationRoles.includes(userRole)) {
    return [{ "$match": { condition } }];
}
.
.
.
condition["organization.name"] = aUserOrganization.orgName;
return [{ "$match": { condition } }];

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let submission = params.submission;
+        const userInfo = context.userInfo;
+        let newApplicationProperties = {
+            _id: v4(),
+            name: params.name,
+            submitterID: userInfo._id,
+            submitterName: formatApplicantName(userInfo),
+            organization: {_id: userInfo.organization.orgID, name: userInfo.organization.orgName},
+            dataCommons: "CDS",
```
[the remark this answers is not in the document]
mtangmt: For this release CDS is the only data commons allowed. I will leave a todo comment

services/submission.js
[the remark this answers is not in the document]
mtangmt: The frontend does not allow this case to happen, but yes, we can add this failsafe.

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    async createSubmission(params, context) {
+        // TODO: Add this requirement: Study abbreviation must have an approved application within user's organization
+        verifySession(context)
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER]);
+        let submission = params.submission;
+        const userInfo = context.userInfo;
+        let newApplicationProperties = {
```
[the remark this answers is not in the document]
mtangmt: This cannot happen beacuse of two failsafes. Frontend cannot send empty string because of form validation. The graphql schema defines these parameters as non-nullable.

services/submission.js
mtangmt: I believe that this is working with Alec's newest changes to user and organization. OrgID should be auto-populated inside the backend by user.js. This may be an issue with your local environment

services/submission.js
mtangmt: The frontend will never send a null or undefined status or organization param. It is defined as such.

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+        const res = await this.submissionCollection.insert(newSubmission);
+        if (!(res?.acknowledged)) {
+            throw new Error(ERROR.CREATE_SUBMISSION_INSERTION_ERROR);
+        }
+        return newSubmission;
+    }
+
+    async listSubmissions(params, context) {
+        verifySession(context)
+            .verifyInitialized();
```
knockknockyoo, line 100: Could you validate submission status? For example, 'NEW' does not return empty.

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+    // List all applications if Fed Lead / Admin / Data Concierge / Data Curator
+    const listAllApplicationRoles = [ROLES.ADMIN, ROLES.FEDERAL_LEAD, ROLES.CURATOR, ROLES.DC_POC];
+    if (listAllApplicationRoles.includes(userRole)) {
+        return [{"$match": conditions}];
+    }
+    // If org owner, add condition to return all data submissions associated with their organization
+    if (userRole === ROLES.ORG_OWNER && aUserOrganization?.orgName) {
+        conditions = {...conditions, "organization.name": aUserOrganization.orgName};
+        return [{"$match": conditions}];
```
knockknockyoo, line 35: Could you double check the requirement? Currently, if you belong to an organization, you wont be able to fetch any submission submitted by submitterID.

services/submission.js
```diff
@@ -0,0 +1,129 @@
...
+
+
+function listConditions(userID, userRole, aUserOrganization, params){
+    const validApplicationStatus = {status: {$in: [NEW, IN_PROGRESS, SUBMITTED, RELEASED, COMPLETED, ARCHIVED]}};
+    // Default conditons are:
+    // Make sure application has valid status
+    let conditions = {...validApplicationStatus};
+    // Filter on organization and status
+    if (params.organization !== ALL_FILTER) {
+        conditions = {...conditions, "organization.name": params.organization};
```
knockknockyoo, line 22: If the organization does not exits, do you think we should throw an error?

knockknockyoo (changes requested):

services/submission.js
mtangmt: Requirements say that org owners don't need to filter by submitterID

services/submission.js
mtangmt: Frontend will only provide organizations from the mongoDB database. So i don't think it is possible for the organization to not exist

knockknockyoo (approved):
Thank you Michale. It looks good to me now.

n2iw (approved):
LGTM.

PR #94, opened 2023-10-10
List submissions poc role fix

mtangmt:
adding the data commons POC filtering.

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -69,10 +69,9 @@ function validateListSubmissionsParams (params) {
 }
 
 class Submission {
-    constructor(logCollection, submissionCollection, organizationCollection) {
+    constructor(logCollection, submissionCollection) {
         this.logCollection = logCollection;
         this.submissionCollection = submissionCollection;
-        this.organizationCollection = organizationCollection;
     }
 
     async createSubmission(params, context) {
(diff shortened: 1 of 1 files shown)


services/submission.js
```diff
@@ -25,13 +25,18 @@ function listConditions(userID, userRole, aUserOrganization, params){
...
+    const listAllApplicationRoles = [ROLES.ADMIN, ROLES.FEDERAL_LEAD, ROLES.CURATOR];
     if (listAllApplicationRoles.includes(userRole)) {
         return [{"$match": conditions}];
     }
-    // If org owner, add condition to return all data submissions associated with their organization
-    if (userRole === ROLES.ORG_OWNER && aUserOrganization?.orgName) {
-        conditions = {...conditions, "organization.name": aUserOrganization.orgName};
+    // If data commons POC, return all data submissions assoicated with their data commons
+    if (userRole === ROLES.DC_POC) {
+        conditions = {...conditions, "dataCommons": userDataCommons};
```
AustinSMueller, line 34: It looks like "userDataCommons" will be determined by context.userInfo.dataCommons. Based on the current GraphQL schema this field is an array and I think filtering on an array of values in MongoDB needs to be handled a little differently. For example:
`{"dataCommons": ["CDS", "ABC"] }` would return no results
`{"dataCommons":  {$in: ["CDS", "ABC"]} }` would return all submissions where the dataCommons = "CDS" or dataCommons = "ABC"

AustinSMueller (approved):

PR #103, opened 2023-10-12
Added getSubmission 

Wesley-Lau:
Added getSubmission

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -199,7 +199,7 @@ class Submission {
         if( Condition_DC_POC || Condition_ORG_OWNER || Condition_SUBMITTER || Condition_Admin){
             return aSubmission
         }else{
-            throw new Error.SUBMISSION_NOT_EXIST
+            throw new Error.INVALID_ROLE
         }    
     }
 }
--- a/services/submission.js
+++ b/services/submission.js
@@ -186,20 +186,20 @@ class Submission {
         verifySession(context)
             .verifyInitialized()
             .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER, ROLES.DC_POC, ROLES.FEDERAL_LEAD, ROLES.CURATOR, ROLES.ADMIN]);
-        let id = params?._id
-        let rUser = await this.userService.getUserByID(context?.userInfo?._id);
+        const id = params?._id
+        const rUser = await this.userService.getUserByID(context?.userInfo?._id);
         const aSubmission = await this.findByID(id);
-        if(context?.userInfo?.role == ROLES.DC_POC){
+        if(context?.userInfo?.role === ROLES.DC_POC){
             if((rUser?.dataCommons.includes(aSubmission?.dataCommons))){
                 return aSubmission
             }
         }
-        if(context?.userInfo?.role == ROLES.ORG_OWNER){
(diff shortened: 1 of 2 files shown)


services/submission.js
```diff
@@ -180,6 +181,33 @@ class Submission {
...
     }
+
+    async getSubmission(params, context){
+        verifySession(context)
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER, ROLES.DC_POC, ROLES.FEDERAL_LEAD, ROLES.CURATOR, ROLES.ADMIN]);
+        let id = params?._id
+        let rUser = await this.userService.getUserByID(context?.userInfo?._id);
+        const aSubmission = await this.findByID(id);
+        if(context?.userInfo?.role == ROLES.DC_POC){
```
AustinSMueller, line 192: We should use the strict equality operator for all of the comparisons in this function. "===" instead of "=="

services/submission.js
```diff
@@ -180,6 +181,33 @@ class Submission {
...
+            if((rUser?.dataCommons.includes(aSubmission?.dataCommons))){
```
AustinSMueller, line 193: instead of nested if statements we can just use a single if statement with an "&&" operator

services/submission.js
```diff
@@ -180,6 +181,33 @@ class Submission {
         }, {"$limit": 1}]);
         return (result?.length > 0) ? result[0] : null;
     }
+
+    async getSubmission(params, context){
```
AustinSMueller, line 185: Rather than have 4 different identical return statements, let's consolidate theses. You can define each acceptance condition as a boolean variable and then use one if block that checks each like in the pseudocode example below. There should also be an error message thrown if the submission is not returned:
```
ownerCondition = (role is org owner AND submission org matches user org)
submitterCondition = (role is submitter AND current user created submission)
...
if (ownerCondition OR submitterCondition OR ...)
    return submission
else
    throw error

**Review by AustinSMueller — CHANGES_REQUESTED** [review]

**Comment by knockknockyoo** [comment]
@Wesley-Lau Could you write the variable names as a camelcase. Let's be consistent. Thanks.

File: services/submission.js [hunk 4]
```diff
@@ -180,6 +181,30 @@ class Submission {
...
+            .verifyInitialized()
+            .verifyRole([ROLES.SUBMITTER, ROLES.ORG_OWNER, ROLES.DC_POC, ROLES.FEDERAL_LEAD, ROLES.CURATOR, ROLES.ADMIN]);
+        const id = params?._id
+        const rUser = await this.userService.getUserByID(context?.userInfo?._id);
+        const aSubmission = await this.findByID(id);
+        if(!aSubmission){
+            throw new Error(ERROR.INVALID_SUBMISSION_NOT_FOUND)
+        }else{
+            // view condition
+            const Condition_DC_POC  = (context?.userInfo?.role === ROLES.DC_POC )&& (rUser?.dataCommons.includes(aSubmission?.dataCommons))
```
knockknockyoo, line 196: Please, use a camelcase. And remove a unnecessary space after Condition_DC_POC. Apply entire variables.

knockknockyoo (approved):

PR #106, opened 2023-10-12
aws bucket name & bug fix

knockknockyoo:
1. submission bucket name changed
2. organization scheme changed
3. organization id already stored in the session


services/submission.js
vshand11, line 1: what's the reason for the following change? is userInfo?.organization?.orgID always works?

knockknockyoo: I've rolled back the change.

vshand11: Approved

vshand11 (approved):
approved

PR #137, opened 2023-10-25
add ccdi data common option

mtangmt:
This pr adds CCDI as an option for data commons parameter when creating a data submission, as required in ticket 525.


knockknockyoo (approved):
It looks good to me. Do you think we should update the comment? "currently only "CDS" is allowed"

AustinSMueller (approved):

PR #143, opened 2023-10-26
update log file dir const

vshand11:
Update the log file dir from log to logs to match default dir in CRDC data loader.


vshand11:
The change been merged in 2.0.0 in the pull request for crdcdh-462.

PR #147, opened 2023-10-30
Update batch sqs msg

knockknockyoo:
1. Added aws SQS service(Standard Option)
- sendSQSMessage
- If AWS FIFO, it requires a groupID
2. Added LOADER_QUEUE as an env
3. Send SQS When
- uploaded successfully all files & metadata


env.template
```diff
@@ -19,6 +19,7 @@ EMAIL_PASSWORD=example password
 EMAIL_URL=http://localhost:XXXX
 EMAILS_ENABLED=true
 SCHEDULE_JOB=1 0 1 * * *
+LOADER_QUEUE=crdcdh-queue
```
vshand11, line 23: For Node.js AWS SDK to load ~/.aws/config, need add an environment variable here as listed below:
AWS_SDK_LOAD_CONFIG=1

vshand11 (changes requested):

vshand11 (approved):
Reviewed and passed testing.

PR #175, opened 2023-11-09
Rootpath Timestamp

knockknockyoo:
Rootpath needs timestamp directory

### DIFF
--- a/services/batch-service.js
+++ b/services/batch-service.js
@@ -133,14 +133,18 @@ const asyncUpdateBatch = async (batchCollection, aBatch) => {
 }
 
 const createPrefix = (params, rootPath, orgID) => {
-    if (rootPath) {
-        return `${rootPath}/${params.type}`;
-    }
-    if (!orgID) {
+    if (!rootPath && !orgID) {
         throw new Error(ERROR.NEW_BATCH_NO_ORGANIZATION);
     }
-    const prefixArray = [orgID, params.submissionID];
-    prefixArray.push(params.type === BATCH.TYPE.METADATA ? BATCH.TYPE.METADATA : BATCH.TYPE.FILE);
+    const prefixArray = rootPath ? [rootPath, params.type] : [orgID, params.submissionID];
+    if (rootPath && params.type === BATCH.TYPE.METADATA) {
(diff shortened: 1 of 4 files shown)


services/batch-service.js
```diff
@@ -134,14 +134,18 @@ const asyncUpdateBatch = async (batchCollection, aBatch) => {
...
-    if (rootPath) {
-        return `${rootPath}/${params.type}`;
-    }
-    if (!orgID) {
+    if (!rootPath && !orgID) {
         throw new Error(ERROR.NEW_BATCH_NO_ORGANIZATION);
     }
-    const prefixArray = [orgID, params.submissionID];
-    prefixArray.push(params.type === BATCH.TYPE.METADATA ? BATCH.TYPE.METADATA : BATCH.TYPE.FILE);
+    const prefixArray = rootPath ? [rootPath, params.type] : [orgID, params.submissionID];
```
n2iw, line 140: I think we should throw and error if rootPath is null, undefined or empty string, because that should not happen under normal circumstances. So we don't need to pass in orgID and try to figure out a default rootPath anymore. In the reset of the function, we know for sure rootPath has valid values.

services/batch-service.js
```diff
@@ -134,14 +134,18 @@ const asyncUpdateBatch = async (batchCollection, aBatch) => {
...
+    if (rootPath && params.type === BATCH.TYPE.METADATA) {
+        // adds timestamp
+        prefixArray.push(getCurrentTime()?.getTime());
```
n2iw, line 143: Instead of get a new timestamp, we should use the batch record's "createdAt" value.

services/batch-service.js
```diff
@@ -134,14 +134,18 @@ const asyncUpdateBatch = async (batchCollection, aBatch) => {
...
+    }
+
+    if (!rootPath && params.type === BATCH.TYPE.FILE) {
```
n2iw, line 146: This should not be needed, since we know rootPath will have values at this point.

n2iw (changes requested):

AustinSMueller (approved):

PR #194, opened 2023-11-29
Add s3 access policy when create temp credential

vshand11:
1) Add s3 access policy when create temp credential.
2) Add errors in submission object in GraphQL schema.

### DIFF
--- a/services/aws-request.js
+++ b/services/aws-request.js
@@ -34,17 +34,28 @@ class AWSService {
         //1. verify token and decode token to get user info
         const userInfo = verifyApiToken(context, config.session_secret);
         //verify submitter
-        await verifySubmitter(userInfo, params?.submissionID, this.submissions, this.userService);
+        const submission = await verifySubmitter(userInfo, params?.submissionID, this.submissions, this.userService);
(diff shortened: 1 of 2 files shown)


services/aws-request.js
```diff
@@ -34,17 +34,28 @@ class AWSService {
...
         const sts = new AWS.STS();
         const timestamp = (new Date()).getTime();
+        //add s3 object access policy
+        const policy = {
+            Version: '2012-10-17',
+            Statement: [
+              {
+                Effect: 'Allow',
+                Action: ['s3:GetObject','s3:PutObject'],
+                Resource: [`arn:aws:s3:::${config.submission_bucket}/${submission.rootPath}/*`]
```
n2iw, line 49: Here we should use **submission.bucketName**, so that if we change how bucket for a submission is determined, we don't have to change code here.

n2iw (changes requested):
Mostly good, except one small change needed.

services/aws-request.js
vshand11: updated.  Please continue.

PR #196, opened 2023-12-05
Added submission level validation results

n2iw:


vshand11 (changes requested):
could you change "validationResult" to "validationStatus", and change "to-be-validated" to "New" that will make the validation status consistent.

n2iw:
> could you change "validationResult" to "validationStatus", and change "to-be-validated" to "New" that will make the validation status consistent.

Agree and changed

PR #199, opened 2023-12-07
CRDCDH-637 

knockknockyoo:
1. created file validation function
2. created a DataSubmission class
3. throw an error if not validation request(roles&validation status)


constants/error-constants.js
```diff
@@ -21,6 +21,7 @@ const ERROR = {
         // submission action
         INVALID_SUBMISSION_ACTION: "Invalid submission action:",
         INVALID_SUBMISSION_ACTION_STATUS: "Invalid submission status for the action:",
+        INVALID_SUBMIT_ACTION: "The submit action should only be allowed if it meets both the user's role and the status of the data submission",
```
AustinSMueller, line 24: Let's reword this to "Invalid submission action, user role and submission status requirements are not met."

verifier/submission-verifier.js
```diff
@@ -41,12 +43,35 @@ class SubmissionActionVerifier {
         this.newStatus = this.actionMap.toStatus;
     }
 
+    isValidSubmitAction(role, aSubmission) {
```
AustinSMueller, line 46: instead of using empty return statements can we flip this logic so that the error is thrown if the conditions are met else it naturally just returns

AustinSMueller (changes requested):

AustinSMueller (approved):

PR #201, opened 2023-12-07
Crdcdh 580

AustinSMueller:


services/data-record-service.js
```diff
@@ -61,6 +66,80 @@ class DataRecordService {
...
+                ROLES.SUBMITTER, // C: can see submission details for his/her own submissions
+                ROLES.DC_POC // D: can see submission details for submissions associated with his/her Data Commons
+            ]);
+        const submissionID = params?._id;
+        const userRole = context.userInfo?.role;
+        let submission = null;
+        if ([ROLES.ORG_OWNER, ROLES.SUBMITTER, ROLES.DC_POC].includes(userRole)){
+            submission = (await this.submissionCollection.find(submissionID)).pop();
+        }
+        if (!!submission && (
```
knockknockyoo, line 85: I think we should move the authentication logic to submission.js. This will let us reuse the submissionQCResults function later on without the authentication. Plus, since submission service is closely linked with datarecords already, it'll help us isolate this function more effectively. Looks good to me. Thanks.

knockknockyoo (changes requested):

services/data-record-service.js
AustinSMueller: I've made these updates to split this function. The authorization logic is now in the submission.js file

knockknockyoo (approved):
Great!

PR #205, opened 2023-12-11
move pagination/sorting out of MongoDB

AustinSMueller:
Since this function uses a custom return type aggregated from several MongoDB objects, the sorting and pagination have to take place after the return objects are created in the backend code

### DIFF
--- a/services/data-record-service.js
+++ b/services/data-record-service.js
@@ -76,19 +76,6 @@ class DataRecordService {
                 }
             }
         });
-        if (!!orderBy){
-            pipeline.push({
-                $sort: {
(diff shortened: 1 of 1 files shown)


services/data-record-service.js
```diff
@@ -112,9 +99,24 @@ class DataRecordService {
                 description: description
             };
         }));
+        if (!!orderBy){
+            const defaultSort = "uploadedDate";
+            const sort = sortDirection === "ASC" ? 1 : -1;
```
knockknockyoo, line 104: can you use the utility 'getSortDirection'?

knockknockyoo (approved):

PR #232, opened 2024-01-08
Added role restriction to listSubmissionNodeTypes

AustinSMueller:


knockknockyoo (commented):
can we remove not operations both conditions? (!!submission && (... and if (await this.#verifyQCResultsReadPermissions(context, submissionID))

AustinSMueller:
> can we remove not operations both conditions? (!!submission && (... and if (await this.#verifyQCResultsReadPermissions(context, submissionID))

I've cleaned up the logic in #verifyQCResultsReadPermissions to make it a little more readable and remove some negations.
I would like to leave the below section the unchanged if that is ok. Generally, I think validation functions should return true if valid, reversing that and making it an invalid check could make it a little more difficult to understand:
```
if (!(await this.#verifyQCResultsReadPermissions(context, submissionID))){
            throw new Error(ERROR.INVALID_PERMISSION_TO_VIEW_NODE_TYPES);
        }
```

knockknockyoo (approved):

PR #234, opened 2024-01-10
CRDCDH-637

knockknockyoo:
1. store&rollback updated time
2. fileValidationStatus is allowed to be null for admin role


vshand11 (approved):
Looks good.

PR #238, opened 2024-01-12
added parameters to submissionQCResults

AustinSMueller:


services/data-record-service.js
```diff
@@ -68,66 +71,85 @@ class DataRecordService {
...
+            $skip: offset
+        });
+        pipeline.push({
+            $limit: first
+        });
+        const dataRecords = await this.dataRecordsCollection.aggregate(pipeline);
         return {
-            total: qcResults.length,
-            results:qcResults.slice(offset, offset+first)
+            total: dataRecords.length,
```
knockknockyoo, line 151: I've faced the same issue. When the **limit** is set, the total count of elements is not the total number of elements in db.

knockknockyoo (changes requested):

services/data-record-service.js
AustinSMueller: I've updated the pipeline to include the correct total

knockknockyoo (approved):
Looks good!

PR #258, opened 2024-02-02
Crdcdh 780

AustinSMueller:


domain/batch.js
```diff
@@ -22,8 +22,16 @@ class Batch {
...
-        const file = new BatchFile(name, size, signedURL, this.filePrefix);
+    addDataFile(name, size){
+        this.#addFile(name, size, null, true)
+    }
+
+    addMetadataFile(name, size, signedURL){
+        this.#addFile(name, size, signedURL, false)
+    }
+
+    #addFile(name, size, signedURL, isDataFile) {
```
knockknockyoo, line 33: can we here signedURL to identify the nodeType instead? I think two dataFile/metadataFile function work as well too.

domain/batch.js
```diff
@@ -45,6 +53,9 @@ class BatchFile {
         this.filePrefix = filePrefix;
         this.createdAt = this.updatedAt = getCurrentTime();
         this.errors = [];
+        if (isDataFile){
```
knockknockyoo, line 56: If isDatafile is false, does that mean we dont intend to create any nodeType?

AustinSMueller: If it is not a data file, then the nodeType is added during validation because it cannot be determined until the file is parsed.

knockknockyoo (approved):

PR #266, opened 2024-02-20
Replace NaN values in submissionQCResults

AustinSMueller:


services/data-record-service.js
```diff
@@ -256,6 +257,17 @@ class DataRecordService {
         });
         return pipeline;
     }
+
+    #replaceNaN(results, replacement){
+        results.map((result) => {
```
knockknockyoo, line 262: results could be null. Could you change it to results?.map?

knockknockyoo (changes requested):

knockknockyoo (approved):

PR #279, opened 2024-03-13
filter submission fileErrors

AustinSMueller:
remove non-object type (non-validation string) errors from the submission.fileErrors and submission.fileWarnings arrays before parsing


n2iw (approved):
LGTM

PR #282, opened 2024-03-19
Update validation submission logics.

vshand11:
The logics for validate both are modified.  Before, validation is sequentially, metadata request is sent first, if failed, return the error so the file validation call is skipped.  After modification,  both metadata and file will be validated then return.


n2iw (approved):
LGTM

PR #291, opened 2024-03-26
If first parameter < 1 then all records are returned

AustinSMueller:


knockknockyoo:
This requires submodule update.

PR #313, opened 2024-04-30
Submission Bucket Origin

knockknockyoo:
replaced the origin of submission bucket name from .env to the field in the submission document.


n2iw (approved):
LGTM

PR #333, opened 2024-05-16
CRDCDH-1046

knockknockyoo:
1. fixed global string replacement
2. email template for inactive submission reminder
3. created cronjob for an inactive submission

### DIFF
--- a/config.js
+++ b/config.js
@@ -8,7 +8,7 @@ let config = {
     inactive_user_days : process.env.INACTIVE_USER_DAYS || 60,
     remind_application_days: process.env.REMIND_APPLICATION_DAYS || 30,
     inactive_application_days : process.env.INACTIVE_APPLICATION_DAYS || 45,
-    remind_submission_days: process.env.INACTIVE_SUBMISSION_DAYS || 60,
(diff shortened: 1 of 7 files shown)


config.js
```diff
@@ -8,6 +8,7 @@ let config = {
     inactive_user_days : process.env.INACTIVE_USER_DAYS || 60,
     remind_application_days: process.env.REMIND_APPLICATION_DAYS || 30,
     inactive_application_days : process.env.INACTIVE_APPLICATION_DAYS || 45,
+    remind_submission_days: process.env.INACTIVE_SUBMISSION_DAYS || 60,
```
n2iw, line 11: Can we change the env var to **INACTIVE_SUBMISSION_NOTIFY_DAYS**, we'll add a new one for deletion days in another ticket.

n2iw (changes requested):

config.js
knockknockyoo: changed applied.

knockknockyoo:
Only submissions in the New, In Progress, Rejected, or Withdrawn status are subject to the inactive notification rule. Submissions in any other state should be ignored.

AustinSMueller (approved):

PR #335, opened 2024-05-22
Implemented crdcdh-1146

vshand11:
Implemented crdcdh-1146

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -618,34 +618,32 @@ class Submission {
         //get target inactive date, current date - config.inactive_submission_days (default 120 days)
         var target_inactive_date = new Date();
         target_inactive_date.setDate(target_inactive_date.getDate() - config.inactive_submission_days);
-        const query = [{"$match": {"status": IN_PROGRESS, "accessedAt": {"$lte": target_inactive_date}}}];
+        const query = [{"$match": {"status": IN_PROGRESS, "accessedAt": {"$exists": true, "$ne": null, "$lte": target_inactive_date}}}];
         try {
             const inactive_subs = await this.submissionCollection.aggregate(query);
             if (!inactive_subs || inactive_subs.length === 0) {
                 console.debug("No inactive submission found.")
                 return "No inactive submissions";
             }
-            const inactive_sub_ids = Array.from(inactive_subs).map(s => s._id);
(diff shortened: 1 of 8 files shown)


services/submission.js
```diff
@@ -610,6 +610,48 @@ class Submission {
...
+        target_inactive_date.setDate(target_inactive_date.getDate() - config.inactive_submission_days);
+        const query = [{"$match": {"status": IN_PROGRESS, "accessedAt": {"$exists": true, "$ne": null, "$lte": target_inactive_date}}}];
+        try {
+            const inactive_subs = await this.submissionCollection.aggregate(query);
+            if (!inactive_subs || inactive_subs.length === 0) {
+                console.debug("No inactive submission found.")
+                return "No inactive submissions";
+            }
+            //delete files under inactive submissions
+            const promises = Array.from(inactive_subs).map(aSub => this.s3Service.deleteDirectory(aSub.bucketName, aSub.rootPath));
```
n2iw, line 629: It expects promise from deleteDirectory, but that function seems doesn't return anything.

n2iw (changes requested):

services/submission.js
vshand11: updated codes not to expect return a promise but wait completion of deletion.   The deleteDirectiroy function is too complicated to implement nested promise since it call itself for deleting dir itself after all objects under dir are deleted.

services/submission.js
```diff
@@ -610,6 +610,44 @@ class Submission {
...
+                console.debug("No inactive submission found.")
+                return "No inactive submissions";
+            }
+            let failed_delete_subs = []
+            //delete related data and files
+            for (const sub of inactive_subs) {
+                try {
+                    const result = await this.s3Service.deleteDirectory(sub.bucketName, sub.rootPath);
+                    if (result === true) {
+                        await this.dataRecordService.deleteMetadataByFilter({"submissionID": sub._id});
```
n2iw, line 634: Let's also delete batches of the submission here.

n2iw (changes requested):

services/submission.js
vshand11: Updated by adding delete related batch.

PR #348, opened 2024-06-11
CRDCDH-1243

knockknockyoo:
This PR will remove only 1 day logic. Instead, it will check inactive submission over 60 days.


AustinSMueller (changes requested):
This looks good, but can you please add the inactiveReminder field to the submission type in the graphql schema?

knockknockyoo:
pushed the inactiveReminder field.

AustinSMueller (approved):

PR #349, opened 2024-06-12
Implemented 1217

vshand11:
List data files.

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -492,7 +492,7 @@ class Submission {
             properties: [],
             nodes: []
         };
-        if (params?.nodeType && params?.nodeType !== "Data File") {
+        if (params?.nodeType !== "Data File") {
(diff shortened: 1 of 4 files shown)


services/submission.js
```diff
@@ -471,35 +471,133 @@ class Submission {
...
-                }
-                node.props = JSON.stringify(node.props);
-                delete node.parents;
-                returnVal.nodes.push(node);
+                returnVal.properties = Array.from(propsSet);
+            }
+            return returnVal;
+            
+        }
+        else {
```
n2iw, line 520: It seems to me the condition of the "else" here is either no "nodeType" or nodeType == "Data File". Is that intention? Should we move condition for nodeType to be inside **if params?.nodeType**?

vshand11: Updated.  Since the nodeType for the query is required so there are only two cases, either "Data File" or not.

n2iw (approved):
LGTM

PR #350, opened 2024-06-13
CRDCDH-1231

knockknockyoo:
1. update the validation properties in the submission document.
2. create a new validation collection & store validation object into validation collection.
3. Send validation id when sending SQS.

### DIFF
--- a/routers/graphql-router.js
+++ b/routers/graphql-router.js
@@ -57,7 +57,7 @@ dbConnector.connect().then(async () => {
     const dataModelInfo = await utilityService.fetchJsonFromUrl(config.model_url);
     const validationCollection = new MongoDBCollection(dbConnector.client, DATABASE_NAME, VALIDATION_COLLECTION);
     const submissionService = new Submission(logCollection, submissionCollection, batchService, userService, organizationService, notificationsService, dataRecordService, config.tier, dataModelInfo, awsService, config.export_queue, s3Service, emailParams, config.dataCommonsList, validationCollection);
-    const dataInterface = new Application(logCollection, applicationCollection, approvedStudiesService, userService, dbService, notificationsService, emailParams, organizationService, config.tier);
+    const dataInterface = new Application(logCollection, applicationCollection, approvedStudiesService, userService, dbService, notificationsService, emailParams, organizationService, config.tier, institutionService);
 
     root = {
         version: () => {return config.version},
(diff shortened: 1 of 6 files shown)


routers/graphql-router.js
```diff
@@ -54,8 +55,9 @@ dbConnector.connect().then(async () => {
...
-    const submissionService = new Submission(logCollection, submissionCollection, batchService, userService, organizationService, notificationsService, dataRecordService, config.tier, dataModelInfo, awsService, config.export_queue, s3Service, emailParams, config.dataCommonsList);
-    const dataInterface = new Application(logCollection, applicationCollection, approvedStudiesService, userService, dbService, notificationsService, emailParams, organizationService, config.tier, institutionService);
+    const validationCollection = new MongoDBCollection(dbConnector.client, DATABASE_NAME, VALIDATION_COLLECTION);
+    const submissionService = new Submission(logCollection, submissionCollection, batchService, userService, organizationService, notificationsService, dataRecordService, config.tier, dataModelInfo, awsService, config.export_queue, s3Service, emailParams, config.dataCommonsList, validationCollection);
+    const dataInterface = new Application(logCollection, applicationCollection, approvedStudiesService, userService, dbService, notificationsService, emailParams, organizationService, config.tier);
```
AustinSMueller, line 60: institutionService is being removed from the Application initialization. I'm guessing this is just a merge error because this is part of the change we just merged but can you make sure to add it back in

services/submission.js
```diff
@@ -1117,6 +1132,41 @@ const isSubmissionPermitted = (aSubmission, userInfo) => {
...
+    constructor(validationType, validationScope, validationStarted) {
+        this.validationStarted = validationStarted ? validationStarted : getCurrentTime();
+        this.validationEnded = null;
+        this.validationType = validationType?.map(type => type.toLowerCase());
+        this.validationScope = validationScope?.toLowerCase();
+    }
+    static createDataValidation(validationType, validationScope, validationStarted) {
+        return new DataValidation(validationType, validationScope, validationStarted);
+    }
+}
```
AustinSMueller, line 1168: Please update the "submission" type in the graphql schema to add these new fields so that they can be queried

services/submission.js
```diff
@@ -1117,6 +1132,41 @@ const isSubmissionPermitted = (aSubmission, userInfo) => {
...
+        this.submissionID = submissionID;
+        this.type = type;
+        this.scope = scope;
+        this.started = getCurrentTime();
+        this.status = status;
+    }
+    static createValidation(submissionID, validationType, validationScope, status) {
+        return new ValidationRecord(submissionID, validationType, validationScope, status);
+    }
+}
```
AustinSMueller, line 1152: Can you add a "Validation" type to the graphql schema? This isn't used but would help document the structure of the validation records and would be useful if we want to implement queries for this in the future

services/submission.js
knockknockyoo: updated

services/submission.js
knockknockyoo: updated

PR #367, opened 2024-07-02
Implemented 1273

vshand11:
1) Implemented crdcdh-1273
2) resolved crdcdh-1321
3) added validation object updating feature if failed to send validation SQS msg.


n2iw (changes requested):
Mostly look good, except the "parents" and "children" return, which I just changed in the ticket to avoid using JSON string. Please refer to the ticket and update the code, thanks!

n2iw (approved):
LGTM

PR #377, opened 2024-07-11
CRDCDH-1288

knockknockyoo:
Update createBatch AP:
for data file, it will generate a fileID

### DIFF
--- a/domain/batch.js
+++ b/domain/batch.js
@@ -1,7 +1,6 @@
 const {getCurrentTime} = require("../crdc-datahub-database-drivers/utility/time-utility");
 const {v4, v5} = require("uuid");
 const {BATCH, FILE} = require("../crdc-datahub-database-drivers/constants/batch-constants");
-const DG_4DFC = "dg.4DFC";
 class Batch {
     constructor(submissionID, displayID, bucketName, filePrefix, type, status) {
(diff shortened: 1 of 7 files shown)


domain/batch.js
```diff
@@ -21,16 +22,25 @@ class Batch {
...
         this.#addFile(name, size, signedURL, false)
     }
 
-    #addFile(name, size, signedURL, isDataFile) {
-        const file = new BatchFile(name, size, signedURL, this.filePrefix, isDataFile);
+    #generateDataFileUUID(fileName, url, dataCommons, studyID) {
+        const urlUUID = v5(url, v5.URL, undefined, undefined);
+        const dataCommonsUUID = v5(dataCommons, urlUUID, undefined, undefined);
+        const studyUUID = v5(studyID, dataCommonsUUID, undefined, undefined);
+        return `${DG_4DFC}/${v5(fileName, studyUUID, undefined, undefined)}`;
```
vshand11, line 38: The prefix "dg.4DFC/" only need to be added if "omit-dcf-prefix" is null or set to false in the datamodel (content.json)

vshand11 (changes requested):

vshand11 (approved):
LGTM

PR #379, opened 2024-07-12
CRDCDH-1261

knockknockyoo:
The status of data record deletion is recorded in the submission.


n2iw (approved):
LGTM. 

FYI. the function **dataRecordService.deleteDataRecords** is quite confusing. It looks like the function is deleting the data instead of send a SQS message. Also, the function probably doesn't belong to dataRecordService, since it's not touching the DB at all.

PR #412, opened 2024-08-21
Implemented new service for CDE

vshand11:
The getCDEs query is required public without authorization.

### DIFF
--- a/routers/api-authorization.js
+++ b/routers/api-authorization.js
@@ -1,0 +1,41 @@
+
+const {parse} = require("graphql");
+
+// escape public query 
+const req_api_name = (req) => {
+    if(req.body && req.body.query){
+        const parsedQuery = parse(req.body.query);
+        const api_name= parsedQuery.definitions.find(
+            (definition) => definition.kind === 'OperationDefinition'
+        ).selectionSet?.selections[0]?.name?.value;
+        return api_name; 
+    }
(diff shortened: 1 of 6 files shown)


routers/api-authorization.js
```diff
@@ -0,0 +1,45 @@
...
+    const fields = schema._queryType?._fields;
+    const public_api_list = Object.values(fields).filter(
+        (f) => (f.astNode.directives.length > 0) && (f.astNode.directives[0].name?.value === api_type)
+      );
+    return public_api_list.map( (api) => api.name )
+}
+
+async function apiAuthorization(req, authenticationService, userInitializationService, public_api_list) {
+    try {
+        if (getAPINameFromReq(req) in public_api_list) return;
```
AustinSMueller, line 28: The array check here needs to be updated. We can use "public_api_list.includes(getAPINameFromReq(req))" The "in" keyword is used to check if an object has a property in javascript 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

AustinSMueller (changes requested):

AustinSMueller (approved):

routers/api-authorization.js
vshand11: Used includes.

PR #431, opened 2024-09-05
Implemented task 1506

vshand11:
Implemented task 1506

### DIFF
--- a/constants/error-constants.js
+++ b/constants/error-constants.js
@@ -123,7 +123,7 @@ const ERROR = {
     // Initialization
     CREATE_USER_MISSING_INFO: "Email and IDP are required to create a new user.",
     CREATE_USER_ORG_MISSING_INFO: "Organization ID is required to initialize the user's organization information.",
-    INVALID_ROLE_STUDY: "No permission the study.",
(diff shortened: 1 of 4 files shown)


constants/error-constants.js
```diff
@@ -123,6 +123,7 @@ const ERROR = {
     // Initialization
     CREATE_USER_MISSING_INFO: "Email and IDP are required to create a new user.",
     CREATE_USER_ORG_MISSING_INFO: "Organization ID is required to initialize the user's organization information.",
+    INVALID_ROLE_STUDY: "No permission the study.",
```
AustinSMueller, line 126: Can we update this to make it clearer that it is a problem with the user's permissions and not a study permission problem? I would recommend "User does not have access to the study" but anything that indicates that it is a user permission problem would be fine

AustinSMueller (changes requested):
Requesting an update to the error message, everything else looks good

constants/error-constants.js
vshand11: Updated, thanks!

AustinSMueller (approved):

PR #442, opened 2024-09-13
validation query optimization

AustinSMueller:
- CRDCDH-1539
- removed batch table join in submissionQCResults and submissionCrossValidationResults pipelines
- moved batch filtering earlier in the pipeline in submissionQCResults and submissionCrossValidationResults pipelines


n2iw (approved):
LGTM

PR #446, opened 2024-09-19
Approved Study

vshand11:
Implemented CRDCDH-1519, 1520 and 1521 for List, Create and Update Approved Study.


routers/graphql-router.js
```diff
@@ -35,6 +35,7 @@ const dbService = new MongoQueries(config.mongo_db_connection_string, DATABASE_N
 const dbConnector = new DatabaseConnector(config.mongo_db_connection_string);
 const AuthenticationService = require("../services/authentication-service");
 const {apiAuthorization, extractAPINames, PUBLIC} = require("./api-authorization");
+const { create } = require('connect-mongo');
```
AustinSMueller, line 38: I think this line can be removed, I don't see this used anywhere

AustinSMueller (approved):

PR #459, opened 2024-09-27
Crdcdh 1519

vshand11:
1. Fixed controlledAccess filter issue.
2. Update sortDirection default to "ASC".
3. Enabled dbGaPID partial match.
4. Make study and dbGaPID match case-insensitive.


n2iw (approved):
LGTM

PR #475, opened 2024-10-08
Crdcdh 1662

vshand11:
Implemented the API editSubmissionCollaborators.


services/submission.js
```diff
@@ -863,6 +863,196 @@ class Submission {
...
+            throw new Error(ERROR.FAILED_ADD_SUBMISSION_COLLABORATOR);
+    }
+
+    /**
+     * API: removeSubmissionCollaborator
+     * @param {*} params 
+     * @param {*} context 
+     * @returns 
+     */
+    async removeSubmissionCollaborator(params, context) {
```
n2iw, line 1029: It seems this function is no longer used?

services/submission.js
```diff
@@ -863,6 +863,196 @@ class Submission {
...
```
n2iw, line 934: Is this function same the one at line 1029?

services/submission.js
vshand11: removed

services/submission.js
vshand11: removed

services/submission.js
```diff
@@ -863,6 +863,130 @@ class Submission {
...
+        const result = await this.submissionCollection.update( aSubmission);
+        if (result?.modifiedCount === 1) {
+            return aSubmission
+        }
+        else
+            throw new Error(ERROR.FAILED_ADD_SUBMISSION_COLLABORATOR);
+    }
+
+    /**
+     * API: editSubmissionCollaborators
```
n2iw, line 929: Looks like this function is same as the one on line 872?

n2iw (changes requested):

n2iw (approved):
LGTM

PR #511, opened 2024-10-23
CRDCDH-1603

knockknockyoo:
The user is attempting to mark the uploaded file as a failed upload.

### DIFF
--- a/constants/error-constants.js
+++ b/constants/error-constants.js
@@ -132,7 +132,7 @@ const ERROR = {
         CONTENTS: "Data file $item$ not found"
     },
     NO_UPLOADED_FILES: "You are attempting to update files that do not exist in the batch. Please check the file: $item$.",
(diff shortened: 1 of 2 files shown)


constants/error-constants.js
```diff
@@ -132,6 +132,7 @@ const ERROR = {
         CONTENTS: "Data file $item$ not found"
     },
     NO_UPLOADED_FILES: "You are attempting to update files that do not exist in the batch. Please check the file: $item$.",
+    INVALID_UPLOAD_ATTEMPT: "You are attempting to falsely update already uploaded files in the s3 bucket. Please check the file: $item$.",
```
AustinSMueller, line 135: I think we need to rewrite this error message, my understanding is that this is failed update to a file that exists. 
My proposed change would be: "Update of existing S3 file $item$ failed. Please check $item$ for errors and then try again."

knockknockyoo: I have updated the error message!

AustinSMueller (approved):

PR #520, opened 2024-10-24
CRDCDH-1721

knockknockyoo:
The requestAccess API can create an organization if the provided organization does not exist.


AustinSMueller (changes requested):
It looks like "submissionCollection", "organizationService", and "appUrl" are no longer used by the UserService in user.js. Please verify and then remove these from the constuctor

knockknockyoo:
Let's get rid of the submodule user soon. Thanks.

AustinSMueller (approved):

PR #533, opened 2024-10-31
CRDCDH-1835

knockknockyoo:
Update createBatch API:
files types to [String!]!
Remove FileInput


AustinSMueller (changes requested):
The FileInput type still needs to be removed from the schema

AustinSMueller (approved):

PR #536, opened 2024-11-05
CRDCDH-1747

knockknockyoo:


services/data-record-service.js
```diff
@@ -827,6 +815,79 @@ class DataRecordService {
...
 
+    #getSubmissionStatQuery(submissionID, validNodeStatus) {
+        return [
+            {$match:{
+                    submissionID: submissionID,
+                    status: {$in: validNodeStatus}
+            }},
+            {$group:{
+                    _id: {submissionID: "$submissionID"},
+                    fileNames: {$addToSet: "$s3FileInfo.fileName"},
```
AustinSMueller, line 826: Since the file names list is split into a separate aggregation, I think the fileNames property here can be removed

AustinSMueller (changes requested):

AustinSMueller (approved):

PR #564, opened 2024-11-27
CRDCDH-2001

knockknockyoo:
When the submission request is in [Inquired, In Progress] status, Fed Leads can submit

### DIFF
--- a/services/application.js
+++ b/services/application.js
@@ -191,11 +191,13 @@ class Application {
 
     async submitApplication(params, context) {
         verifySession(context)
-            .verifyInitialized();
+            .verifyInitialized()
+            .verifyRole([USER.ROLES.SUBMITTER, USER.ROLES.FEDERAL_LEAD])
         const application = await this.getApplicationById(params._id);
         verifyApplication(application)
             .notEmpty()
-            .state([NEW, IN_PROGRESS]);
+            // Submitter or Fed Leads only can submit under certain application status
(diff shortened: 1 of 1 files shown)


services/application.js
```diff
@@ -191,11 +191,13 @@ class Application {
...
         verifySession(context)
-            .verifyInitialized();
+            .verifyInitialized()
+            .verifyRole([USER.ROLES.SUBMITTER, USER.ROLES.FEDERAL_LEAD])
         const application = await this.getApplicationById(params._id);
         verifyApplication(application)
             .notEmpty()
-            .state([NEW, IN_PROGRESS]);
+            // Submitter or Fed Leads only can submit under certain application status
+            .state(context?.userInfo?.role === USER.ROLES.SUBMITTER ? [NEW, IN_PROGRESS] : [INQUIRED, IN_PROGRESS]);
```
AustinSMueller, line 200: I know that with the current code the only possible roles at this line are SUBMITTER and FEDERAL_LEAD but I think we should do a role check anyway to prevent issues resulting from any future changes. If any non-submitters reach this state they will have the same submit permissions as FEDERAL_LEADS

AustinSMueller (changes requested):

services/application.js
knockknockyoo: I've pushed the change. The empty status will be assigned for invalid roles.

AustinSMueller (approved):

PR #572, opened 2024-12-05
CRDCDH-2010

vshand11:
Implemented 1) Remove listApprovedStudiesOfMyOrganization API
    2) Update createSubmission to verify if studyID parameter is within a user's studies property instead of through organization.


n2iw (approved):
LGTM

PR #586, opened 2024-12-12
CRDCDH-2017

knockknockyoo:
Update editUser to API
- Add a permissions
- Add a notifications


services/user.js
```diff
@@ -741,6 +748,57 @@ class UserService {
         });
         return await this.userCollection.aggregate(pipeline);
     }
+
+    #validateUserPermission(permissions, notifications) {
+        // Only Valid User Permissions
+        if (permissions) {
+            const allPermissionNames = [...Object.values(SUBMISSION_REQUEST), ...Object.values(DATA_SUBMISSION), ...Object.values(ADMIN)];
+            const allPermissionNamesSet = new Set(allPermissionNames);
```
AustinSMueller, line 756: allPermissionNameSet and allEmailNotificationNamesSet are based on constants so they can be defined at initialization instead of each time this function is called

services/user.js
```diff
@@ -803,7 +861,144 @@ class DataCommon {
...
+                // For USER.ROLES.USER:
+                return [SUBMISSION_REQUEST.CREATE, DATA_SUBMISSION.REQUEST_ACCESS];
+        }
+    }
+
+    #getBannedPermissionsByRole() {
+        switch (this.role) {
+            case USER.ROLES.FEDERAL_LEAD:
+                return [DATA_SUBMISSION.ADMIN_SUBMIT, DATA_SUBMISSION.REQUEST_ACCESS];
+            case USER.ROLES.DATA_COMMONS_PERSONNEL:
```
AustinSMueller, line 911: There are only 2 banned permissions for Data commons personnel based on the latest spreadsheet for the ticket

AustinSMueller (changes requested):

services/user.js
knockknockyoo: I've pushed it for the change.

services/user.js
knockknockyoo: Update it for the change.

PR #594, opened 2024-12-18
CRDCDH-2121

vshand11:
Implemented CRDCDH-2121

### DIFF
--- a/services/configurationService.js
+++ b/services/configurationService.js
@@ -1,3 +1,7 @@
+const {verifySession} = require("../verifier/user-info-verifier");
+const USER_CONSTANTS = require("../crdc-datahub-database-drivers/constants/user-constants");
+const ROLES = USER_CONSTANTS.USER.ROLES;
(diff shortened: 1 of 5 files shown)


services/configurationService.js
```diff
@@ -9,6 +13,33 @@ class ConfigurationService {
...
+            .verifyRole([ROLES.ADMIN]);
+
+        return await this.getPBACByRoles(params.roles);
+    }
+    /**
+     * Get PBAC defaults by roles
+     * @param {Array} roles
+     * @returns {Object} PBAC defaults
+     */
+    async getPBACByRoles(roles){
```
knockknockyoo, line 34: Can you set it private #getPBACByRoles?

knockknockyoo (changes requested):
Please, see the comment.

services/configurationService.js
vshand11: It's not private, other service will use this function.

knockknockyoo (approved):

PR #599, opened 2024-12-24
CRDCDH-2170

vshand11:
Rollback the recent change made by Young 2 month ago that cause the issue in prod only in CRDCDH-2170 in extract tier name function in config.js

### DIFF
--- a/config.js
+++ b/config.js
@@ -102,7 +102,9 @@ function getModelUrl() {
 }
 
 function extractTierName() {
-    return process.env.TIER?.replace(/prod(uction)?/gi, '')?.replace(/[^a-zA-Z\d]/g, '')?.trim();
+    // return process.env.TIER?.replace(/prod(uction)?/gi, '')?.replace(/[^a-zA-Z\d]/g, '')?.trim();
+    // rollback the change to fix CRDCDH-2170
+    return process.env.TIER?.replace(/[^a-zA-Z\d]/g, '')?.trim();
 }
 
 function getTier() {
--- a/config.js
+++ b/config.js
@@ -92,7 +92,7 @@ function getModelUrl() {
     if (process.env.MODEL_URL) {
         return process.env.MODEL_URL;
     }
-    const tier = extractTierName();
+    const tier = process.env.TIER?.replace(/[^a-zA-Z\d]/g, '')?.trim();
     // By default url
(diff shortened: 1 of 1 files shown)


knockknockyoo (changes requested):
We need this code to generate the tier name(empty) on production.

vshand11:
> We need this code to generate the tier name(empty) on production.

If the tier is changed to empty sting in production, the data model url will point to master branch and the latest version for cds in master branch is "4.0.4".  This is the issue reported in the ticket.  In production, it should point to "prod" branch so the code need to be removed.

vshand11:
I see

vshand11:
> > > We need this code to generate the tier name(empty) on production.
> > 
> > 
> > If the tier is changed to empty sting in production, the data model url will point to master branch and the latest version for cds in master branch is "4.0.4". This is the issue reported in the ticket. In production, it should point to "prod" branch so the code need to be removed.
> 
> We can't remove the code. If we remove the code, it will cause other issue on email notification template. We need the different approach.

Updated.  Please continue.

config.js
```diff
@@ -92,7 +92,7 @@ function getModelUrl() {
     if (process.env.MODEL_URL) {
         return process.env.MODEL_URL;
     }
-    const tier = extractTierName();
+    const tier = process.env.TIER?.replace(/[^a-zA-Z\d]/g, '')?.trim();
```
knockknockyoo, line 95: On production, we've encountered issues with the DevOps team hardcoding 'prod' or 'production.' If these values are set in the env, the same issue could arise. If the value is empty, should we default to **prod** instead of master?

I know this is a very tricky situation. Hope you understand. 
['https://raw.githubusercontent.com/CBIIT/crdc-datahub-models/', tier || '**prod**', '/cache/content.json']

vshand11: For data model URL,  we don't need to make a special logic for production other then others.   The environment variable TIER is set to "prod",  and BE 3.1.0 uses the env setting for tier.   If set default to "prod", it may cause issue if the environment variable TIER was not set for a new tier for example.

knockknockyoo: I will merge the PR. But, we need to make sure what the env variable is set on production.

knockknockyoo (approved):

PR #601, opened 2024-12-30
CRDCDH-2017

knockknockyoo:


services/configurationService.js
```diff
@@ -40,8 +40,57 @@ class ConfigurationService {
...
+
+    /**
+     * Get User Permission Access by a role
+     * UserAccessControl class to filter permissions.
+     * @param {String} role - a user role.
+     * @returns {Promise<{ permissions: {disabled: Array, permitted: Array}, notifications: {disabled: Array, permitted: Array} }> | null}
+     */
+    async getAccessControl(role) {
+        const usersPermissions = await this.getPBACByRoles([role]);
+        if (usersPermissions.length === 0) {
```
AustinSMueller, line 52: please make this length check null safe

AustinSMueller (changes requested):

services/configurationService.js
knockknockyoo: null safe check added

AustinSMueller (approved):

PR #602, opened 2024-12-30
CRDCDH-2093

vshand11:
Implemented CRDCDH-2093.

### DIFF
--- a/services/user.js
+++ b/services/user.js
@@ -16,7 +16,6 @@ const {
     EMAIL_NOTIFICATIONS: EN,
 } = require("../crdc-datahub-database-drivers/constants/user-permission-constants");
 
-
 const isLoggedInOrThrow = (context) => {
     if (!context?.userInfo?.email || !context?.userInfo?.IDP) throw new Error(SUBMODULE_ERROR.NOT_LOGGED_IN);
 }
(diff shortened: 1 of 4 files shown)


services/user.js
```diff
@@ -793,6 +793,17 @@ class UserService {
...
         }
     }
+
+    async getUsersByStudyID(studyID) {
+        const query = {
+            "studies": {"$in": [studyID, "All"]}
+        }; // user's studies contains studyID
+        const users = await this.userCollection.aggregate([{"$match": query}]);
+        for (const user of users) {
+            user.studies = await this.#findApprovedStudies(user.studies);
```
knockknockyoo, line 803: Can we collect all approved IDs from the users and make a single query using **await this.#findApprovedStudies**? Querying for all users could lead to performance issues.

knockknockyoo (changes requested):

services/user.js
vshand11: Since each user has different studies, need retrieve studies one user by one user.

knockknockyoo: We can start by collecting all the study IDs from the requested users. Then, perform a single query using those study IDs. The results will include these study IDs, which can then be iterated over in the loop for further processing. Is that possible? I think this might solve the performance issue.

vshand11: It's possible but make the code very complex.  I'll use a special query rather than getUserByStudy

vshand11:
Optimized the query logic to limited users returned.

knockknockyoo (approved):

services/user.js
```diff
@@ -794,7 +794,8 @@ class UserService {
     } 
     async getCollaboratorsByStudyID(studyID, submitterID) {
         const query = {
-            "studies": {"$in": [studyID, "All"]}, "role": USER.ROLES.SUBMITTER, _id: {"$ne": submitterID}
+            "role": USER.ROLES.SUBMITTER, _id: {"$ne": submitterID},
+            "$or": [{"studies": {"$in": [studyID, "All"]}}, {"studies._id": {"$in": [studyID, "All"]}}]
```
knockknockyoo, line 798: Are we supposed to store only study ids for the studies? This might lead some confusion.

vshand11: In user collection,  the studies property is array of object {_id: String} or array of string (backward compatible).

AustinSMueller (approved):

PR #625, opened 2025-01-08
CRDCDH-2138

knockknockyoo:


constants/error-constants.js
```diff
@@ -112,6 +112,7 @@ const ERROR = {
     // delete submission error file list
     DELETE_NO_FILE_SUBMISSION: "No extra file found",
     DELETE_NO_DATA_FILE_EXISTS: "No data files found for deletion in the bucket",
+    INVALID_DELETE_SUBMISSION_STATUS : "Requested submission is not allowed to delete",
```
AustinSMueller, line 115: Let's update this message to "Data Submissions that have been released cannot be deleted"

AustinSMueller (changes requested):

constants/error-constants.js
knockknockyoo: pushed it for the update.

AustinSMueller (approved):

PR #633, opened 2025-01-09
Add order property in both Permission and Notification types

vshand11:
Add order property in both Permission and Notification types per FE request.


vshand11:
Closed the PR since Alec made changes.

PR #636, opened 2025-01-09
CRDCDH-2166

knockknockyoo:


constants/error-constants.js
```diff
@@ -7,7 +7,9 @@ const ERROR = {
     FAILED_DELETE_APPLICATION: "An error occurred while delete an application",
     APPLICATION_CONTROLLED_ACCESS_NOT_FOUND: "The application does not store controlled access property.",
     APPLICATION_INVALID_STATUES: "Requested statues $item$ is not valid",
+    FAILED_RESTORE_APPLICATION: "An error occurred while restoring an application",
     MISSING_PROGRAM_INFO: "The program property is required to approve the submission request.",
+    INVALID_APPLICATION_RESTORE_STATE: "Application is failed to restore because of the invalid previous history",
```
AustinSMueller, line 12: Let's reword this to: "Failed to restore application because the application history data is invalid"

AustinSMueller (changes requested):

constants/error-constants.js
knockknockyoo: Updated the error message.

AustinSMueller (approved):

PR #637, opened 2025-01-09
Updated verifyPermission

vshand11:
Updated verifyPermission to support both permission and array of permissions


knockknockyoo (approved):

knockknockyoo (approved):
For future reference, I would like to add the comment for the function: "When at least one element meets the condition, it will pass."

AustinSMueller (approved):

PR #643, opened 2025-01-14
Updated submit SR

vshand11:
Updated the logic of valid status in submitApplication API based on new requirement.

### DIFF
--- a/services/application.js
+++ b/services/application.js
@@ -198,12 +198,7 @@ class Application {
             .verifyInitialized()
             .verifyPermission(USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.SUBMIT);
         const application = await this.getApplicationById(params._id);
-        let validStatus = [];
-        if (context?.userInfo?.role === USER.ROLES.SUBMITTER) {
-            validStatus = [NEW, IN_PROGRESS];
-        } else if (context?.userInfo?.role === USER.ROLES.FEDERAL_LEAD) {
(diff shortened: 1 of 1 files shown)


services/application.js
```diff
@@ -198,12 +198,7 @@ class Application {
             .verifyInitialized()
             .verifyPermission(USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.SUBMIT);
         const application = await this.getApplicationById(params._id);
-        let validStatus = [];
-        if (context?.userInfo?.role === USER.ROLES.SUBMITTER) {
-            validStatus = [NEW, IN_PROGRESS];
-        } else if (context?.userInfo?.role === USER.ROLES.FEDERAL_LEAD) {
-            validStatus = [INQUIRED, IN_PROGRESS];
-        }
+        const validStatus = [IN_PROGRESS, INQUIRED]; //updated based on new requirement.
```
knockknockyoo, line 201: Can I see the user-story related to this?

n2iw: Young, there is no user story specific about this, but as we moved to PBAC, this status check should not be based on roles anymore.

n2iw (approved):
LGTM

services/application.js
n2iw: > Can I see the user-story related to this? I can't find.

Young, there is no user story specific about this, but as we moved to PBAC, this status check should not be based on roles anymore.

knockknockyoo: Thanks.

PR #644, opened 2025-01-14
CRDCDH 2093 update edit collaborator

vshand11:
update editSubmissionCollaborators API to verify user's study directly.


services/submission.js
```diff
@@ -996,16 +996,14 @@ class Submission {
...
+                if(!user?.studies || user.studies.length === 0 )
                     throw new Error(ERROR.INVALID_COLLABORATOR_STUDY);
-                }
-                const collaborator_study = organization.studies.find(s => s._id ===  aSubmission.studyID);
-                if (!collaborator_study)
-                {
+                const userStudy = (user.studies[0] instanceof Object)? user.studies.find(s=>s._id === aSubmission.studyID || s._id === "All"):
+                    user.studies.find(s=> s === aSubmission.studyID || s === "All"); //backward compatible
+
+                if (!userStudy)
```
knockknockyoo, line 1005: Peter, could you create this logic as a function. This logic requires a lot of APIs.

knockknockyoo:
The requirement in your ticket. This is missing. Can you check?
`and batch service need to be removed.`

services/submission.js
vshand11: Updated.

vshand11:
> The requirement in your ticket. This is missing. Can you check? `and batch service need to be removed.`

The "Can View" has been removed from batch service.

n2iw (approved):
LGTM

PR #648, opened 2025-01-16
Updated Admin default permission based on latest user story CRDCDH-1888

n2iw:


vshand11 (approved):
Compared one by one with the US.  Matched

PR #651, opened 2025-01-17
regarding the user-control issue.

knockknockyoo:


services/user.js
```diff
@@ -60,6 +60,7 @@ class UserService {
             .verifyPermission(USER_PERMISSION_CONSTANTS.DATA_SUBMISSION.REQUEST_ACCESS);
 
         // USER.ROLES.ORG_OWNER needs to be removed after the role is retired finally
+        // TODO
         if (![USER.ROLES.SUBMITTER, USER.ROLES.USER, USER.ROLES.ORG_OWNER].includes(params.role)) {
```
n2iw, line 64: This role limitation should be removed.

services/user.js
```diff
@@ -428,12 +429,13 @@ class UserService {
             throw new Error(SUBMODULE_ERROR.USER_NOT_FOUND);
         }
         const updatedUser = {};
+        // TODO
         const isCurator = updatedUser?.role === USER.ROLES.CURATOR || user[0]?.role === USER.ROLES.CURATOR || params?.role === USER.ROLES.CURATOR;
```
n2iw, line 433: Seems isCurator is only used later in line 452 to set organization to null, I don't think it's still needed. So I suggest we remove this line and line 452-454

services/user.js
```diff
@@ -428,12 +429,13 @@ class UserService {
...
 
         if (params.role && Object.values(USER.ROLES).includes(params.role)) {
             updatedUser.role = params.role;
         }
-
+        // TODO
         if(!params?.studies && ![USER.ROLES.ADMIN, USER.ROLES.USER, USER.ROLES.CURATOR, USER.ROLES.DC_POC, USER.ROLES.DATA_COMMONS_PERSONNEL].includes(params.role)){
```
n2iw, line 439: This should be changed to only require Submitter role to have study assigned (don't use reverse logic)

n2iw (changes requested):

knockknockyoo (commented):
Please, re-review. Thank you.

services/user.js
knockknockyoo: I've updated the change.

services/user.js
knockknockyoo: I've updated the change.

n2iw (approved):
LGTM

PR #657, opened 2025-01-21
Remove User Organization Check

AustinSMueller:
Remove User Organization check for when a user is creating a submission. User organization is not used anymore


knockknockyoo (approved):
Approved

PR #665, opened 2025-01-24
Update admin defaults based on updated story CRDCDH-2111

n2iw:


resources/json/PBACDefaults_config.json
```diff
@@ -691,7 +691,7 @@
             "group": "Submission Request Emails",
             "name": "When deleted",
             "order": 6,
-            "checked": true,
+            "checked": false,
```
AustinSMueller, line 694: This looks like it should be true based on the PBAC spreadsheet

vshand11 (approved):
Changes match with US 2111

resources/json/PBACDefaults_config.json
AustinSMueller: This matches the user story, the PBAC spreadsheet is incorrect

PR #666, opened 2025-01-24
CRDCDH 2253

vshand11:
1) Add new data field, version, for application.
2) Add new configuration, APPLICATION_FORM_VERSIONS, into DB, please ask me to add the record when deploying to dev2.
3) Updated relatedAPIs to set the version if not stored in Application docs.


services/application.js
```diff
@@ -40,9 +41,22 @@ class Application {
...
         }
+        // populate the version if not stored in
+        if (!application?.version){
+            application.version  = await this.#getApplicationVersionByStatus(application.status);
+        }
         return application;
     }
 
+    #getApplicationVersionByStatus(status) {
+        const config = this.configurationService.findByType("APPLICATION_FORM_VERSIONS");//get the default form versions dynamically
```
knockknockyoo, line 52: This should be the async function.

knockknockyoo:
Can we run a query to update the existing submission request to include the version property instead of writing codes? I believe we should only include the version in the createApplication function.

services/application.js
vshand11: Good catch, updated.   Thanks!

vshand11:
> Can we run a query to update the existing submission request to include the version property instead of writing codes? I believe we should only include the version in the createApplication function.

There should be another task to update the existing data.   BE codes need make sure if the version is not stored, we need return the version to FE.

knockknockyoo:
> > Can we run a query to update the existing submission request to include the version property instead of writing codes? I believe we should only include the version in the createApplication function.
> 
> There should be another task to update the existing data. BE codes need make sure if the version is not stored, we need return the version to FE.

After the query executed, the version should be stored in the submission request. Then, the migration script should be completed before the PR.

vshand11:
> > > Can we run a query to update the existing submission request to include the version property instead of writing codes? I believe we should only include the version in the createApplication function.
> > 
> > 
> > There should be another task to update the existing data. BE codes need make sure if the version is not stored, we need return the version to FE.
> 
> After the query executed, the version should be stored in the submission request. Then, the migration script should be completed before the PR.

Yes, but BE codes still need to implemented the logic in case the version is not stored that makes the BE bullet-proof.

knockknockyoo:
> > > > Can we run a query to update the existing submission request to include the version property instead of writing codes? I believe we should only include the version in the createApplication function.
> > > 
> > > 
> > > There should be another task to update the existing data. BE codes need make sure if the version is not stored, we need return the version to FE.
> > 
> > 
> > After the query executed, the version should be stored in the submission request. Then, the migration script should be completed before the PR.
> 
> Yes, but BE codes still need to implemented the logic in case the version is not stored that makes the BE bullet-proof.

It is up to you. The version is stored in the database.

knockknockyoo (approved):

PR #671, opened 2025-01-28
CRDCDH-2251

knockknockyoo:


services/submission.js
```diff
@@ -2072,6 +2062,33 @@ class Collaborators {
     }
 }
 
+// TODO remove temporary for QA
```
AustinSMueller, line 2065: Do there need to be removed before merging?

AustinSMueller, line 2065: Do there need to be removed before merging?

AustinSMueller (changes requested):

services/submission.js
knockknockyoo: QA testing purpose. We will remove it.

AustinSMueller (approved):

PR #678, opened 2025-01-30
Crdcdh 2252

vshand11:
implement a cron job to purge to-be-delete folder after "Delete" type of Data Submissions are completed 180 days.

- add a configuration for purge days in DB.
- created a function to purge deleted files 180 after created with the completed tag.
- create a cron job to run the function every night.


services/submission.js
```diff
@@ -1215,8 +1217,27 @@ class Submission {
...
+     * purgeDeletedDataFiles
+     * remove deleted data files in "to-be-deleted" after tagged with "Completed = true"
+     */
+    async purgeDeletedDataFiles(){
+        //get target purge date, current date - config.purgeDeletedDataFileDays (default 180 days)
+        const purgeConfig = await this.configurationService.findByType("PURGE_DELETED_DATA_FILE");
+        const purgeDays = purgeConfig?.days ?? 180;
+        const folder = purgeConfig?.prefix ?? "to_be_deleted";
+        const tag = purgeConfig?.tag ?? {key: "Completed", value: "true"};
+        const dmBucketConfig = await this.configurationService.findByType("DM_BUCKET_NAME");
```
knockknockyoo, line 1230: Can we declare this as private constant? #DM_BUCKET_NAME= "DM_BUCKET_NAME"

services/submission.js
```diff
@@ -1215,8 +1217,27 @@ class Submission {
...
+        const dmBucketName = dmBucketConfig?.keys.dm_bucket;
+        try {
 
-
+            await this.s3Service.purgeDeletedFiles(dmBucketName, folder, purgeDays, tag);
+            console.debug("Successfully purged deleted data files."); 
+        }
+        catch (e){
+            console.error("Failed to purge deleted data files with error:" + e.message);
```
knockknockyoo, line 1238: I think we should leave some sources, like dmBucketName, to better track errors.

services/submission.js
vshand11: Since the "DM_BUCKET_NAME" is only used once, there are no benefits to add a constant.

services/s3-service.js
```diff
@@ -0,0 +1,227 @@
...
+            console.info("No data files to be purged.");
+        return true;  // if no exceptions
+    }
+
+    async #deleteObjects(bucketName, fileKeyList) {
+        const deleteParams = {
+            Bucket: bucketName,
+            Delete: { Objects: fileKeyList.map(fileKey => ({ Key: fileKey})) }
+        };
+        await this.s3.deleteObjects(deleteParams).promise();
```
knockknockyoo, line 221: I think we should verify that the files are deleted. If failed, we should log error messages via deleteObjects. Additionally, we should consider renaming the function to 'deleteDataFileObjects'?

services/submission.js
vshand11: Updated.

services/s3-service.js
vshand11: If there are any errors, the caller in submission will catch exception and printed the error.  Since the private function is generic,  the name is ok to me.

knockknockyoo: Got it.

knockknockyoo (approved):

n2iw (approved):
LGTM

PR #705, opened 2025-02-06
CRDCDH-2322

knockknockyoo:


services/submission.js
```diff
@@ -1654,7 +1654,7 @@ const sendEmails = {
                     submissionName: `${aSubmission?.name},`,
                     dataCommonsName: aSubmission?.dataCommons,
                     contactName: `${aSubmission?.conciergeName || 'NA'}`,
-                    contactEmail: `${aSubmission?.contactEmail || 'NA'}.`
+                    contactEmail: `${aSubmission?.conciergeName || 'NA'}.`
```
AustinSMueller, line 1657: Why is conciergeName used for both of these fields here? Is this correct?

AustinSMueller (changes requested):

services/submission.js
knockknockyoo: My apology. This should be conciergeEmail. I have pushed the change.

PR #711, opened 2025-02-10
CRDCDH-2338

vshand11:
Implemented CRDCDH-2338 to send a SQS message when a released submission with "Delete" intention is rejected.

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -425,7 +425,8 @@ class Submission {
             completePromise.push(this.dataRecordService.exportMetadata(submissionID));
             completePromise.push(this.#sendCompleteMessage({type: GENERATE_DCF_MANIFEST, submissionID}, submissionID));
         }
-        if (action === ACTIONS.REJECT) {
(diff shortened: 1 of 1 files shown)


knockknockyoo (approved):
Apprved!

services/submission.js
```diff
@@ -35,6 +35,7 @@ const DATA_MODEL_SEMANTICS = 'semantics';
 const DATA_MODEL_FILE_NODES = 'file-nodes';
 const COMPLETE_SUBMISSION = "Complete Submission";
 const GENERATE_DCF_MANIFEST = "Generate DCF Manifest";
+const RESTORE_DELETED_DATA_FILE = "Restore Deleted Data File";
```
n2iw, line 38: It will be better to say "Files" (plural)

n2iw (changes requested):

services/submission.js
vshand11: updated

PR #714, opened 2025-02-11
CRDCDH 2161

vshand11:
Updated based on changes in user story and task as listed below:
-Update createApprovedStudy and updateApprovedStudy APIs and make primaryContactID parameter optional 
-Update ApprovedStudy GraphQL type and all APIs which returns that type and change program to programs and make is [UserOrganization]

### DIFF
--- a/services/approved-studies.js
+++ b/services/approved-studies.js
@@ -75,8 +75,8 @@ class ApprovedStudiesService {
         }
         let returnStudy = study[0];
         // find program/organization by study ID
-        returnStudy.program = await this.#findOrganizationByStudyID(_id)
-        // file primaryContact
+        returnStudy.programs = await this.#findOrganizationByStudyID(_id)
+        // find primaryContact
         if (returnStudy?.primaryContactID)
         {
             returnStudy.primaryContact = await this.#findUserByID(returnStudy.primaryContactID);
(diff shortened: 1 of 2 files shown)


services/approved-studies.js
```diff
@@ -219,12 +220,15 @@ class ApprovedStudiesService {
...
         // check primaryContactID 
-        const primaryContact = await this.#findUserByID(primaryContactID);
-        if (!primaryContact) {
-            throw new Error(ERROR.INVALID_PRIMARY_CONTACT);
-        }
-        if (primaryContact.role !== USER.ROLES.DATA_COMMONS_PERSONNEL){
-            throw new Error(ERROR.INVALID_PRIMARY_CONTACT_ROLE);
+        let primaryContact = null;
+        if (!primaryContactID) {
+            const primaryContact = await this.#findUserByID(primaryContactID);
```
knockknockyoo, line 225: This would be not ideal **primaryContact** already declared null;
let => const?

vshand11: It can't be const because if could be reassigned to what find by the primaryContactID.   It will be add to return value.

knockknockyoo: The value is changing let => const
Please, take a look at the line 293.
primaryContact = await this.#findUserByID(primaryContactID);

vshand11: Oh,  updated, thanks!

knockknockyoo (approved):
Perfect!

AustinSMueller (approved):

PR #716, opened 2025-02-12
CRDCDH 2253

vshand11:
Updated based on user story 2284 to auto upgrade the version when save.

### DIFF
--- a/services/application.js
+++ b/services/application.js
@@ -42,21 +42,18 @@ class Application {
         if (application.status === APPROVED){
             await this.#checkConditionalApproval(application);
         }
-        // populate the version if not stored in
-        if (!application?.version){
-            application.version  = await this.#getApplicationVersionByStatus(application.status);
-        }
+        // populate the version with auto upgrade based on configuration
+        application.version  = await this.#getApplicationVersionByStatus(application.status, application.version);
         return application;
     }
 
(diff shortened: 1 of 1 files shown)


services/application.js
```diff
@@ -42,21 +42,18 @@ class Application {
...
-        const defaultVersion = "1.0";
-        const newVersion = "2.0"; 
+    async #getApplicationVersionByStatus(status, version = null ) {   
         const config = await this.configurationService.findByType("APPLICATION_FORM_VERSIONS"); //get version config dynamically
-        const currentVersion = config?.current || defaultVersion;
-        const newStatusVersion = config?.new || newVersion;
-
-        return [NEW, IN_PROGRESS, INQUIRED].includes(status) ? newStatusVersion : currentVersion;
+        const currentVersion = config?.current || "2.0";
+        const newStatusVersion = config?.new || "3.0";
```
knockknockyoo, line 53: In my opinion, any version should not be included in the codes. The version will be changed anyway. We should create the migration script instead. I will leave it up to you.

knockknockyoo (approved):

services/application.js
vshand11: After restore, the version will be changed,  so the version need to be upgraded when save a application.  this is required by the user story 2284.

AustinSMueller (approved):

PR #731, opened 2025-03-05
Fix severity filter (1637)

AustinSMueller:
- updates to validation results severity filter
- fix misplaced quotation in "Set Federal Leads Permissions" command


knockknockyoo (approved):
Perfect!

PR #745, opened 2025-03-17
CRDCDH 2436

vshand11:
1. add new configuration in DB, UPLOADING_HEARTBEAT,
2. updated updateBatch API definition by adding a new param, uploading: boolean, for heartbeat call from CLI.
3. added new class UploadingMonitor in batch-services.js
4. updated updateBatch function in submission.js to use the UploadingMonitor.
5. updated graphql-router.js to use the UploadingMonitor.
6. add new errors and constants.


knockknockyoo (approved):
Approved!

AustinSMueller (approved):

PR #771, opened 2025-04-03
CRDCDH-2528

knockknockyoo:


services/approved-studies.js
```diff
@@ -329,10 +330,25 @@ class ApprovedStudiesService {
...
+
         return {...updateStudy, programs: programs, primaryContact: primaryContact};  
     }
+
+    #getConcierge(programs, primaryContact) {
+        //update conciergeName and conciergeEmail in submissions no matter if primaryContact is null or not
+        const [conciergeName, conciergeEmail] = (primaryContact)? [`${primaryContact?.firstName || ""} ${primaryContact?.lastName || ''}`, primaryContact?.email || ""] :
+            ["",""];
+
+        // Study doesn't have a primary contact, so look for the primary contact in the program instead
```
vshand11, line 342: If study don't have primary contact, set conciergeName, conciergeEmail to "".

vshand11 (changes requested):

services/approved-studies.js
knockknockyoo: This is what exactly setting conciergeName/conciergeEmail ["",""]. What changes are you referring to?

vshand11: The codes below still trying to set conciergeName/conciergeEmail based on program.

knockknockyoo: Yes, if the primary contact is not available from the study, the primary contact should be assigned from the program.

vshand11: This is not correct logic now.

vshand11: The requirement is set conciergeName/conciergeEmail only based on study.primaryContactID, not program anymore.

knockknockyoo: When the user unchecks the primary contact in the UI, it means they're no longer the primary contact in the study. It should assigns the primary contact from the program. Did I misunderstand the process by changing the approvedStudies API? 
Screenshot 2025-04-03 at 11 56 45 AM

knockknockyoo: I believe **Same as the Program Primary Contact** checked, conciergeName/conciergeEmail with the primary contact in the program for the existing data-submission should be assigned. In the ticket, new data submission is correctly stored described in the ticket.

vshand11: If you check discussions in developers channel last two weeks,  both Ming and Alec emphasized set primaryContactID to null if user clicked Same as the Program Primary Contact.   You may double check either of them.  If uncheck, user needs select primary Contact from a list.

knockknockyoo: Okay. I will bring up this.

knockknockyoo: The change has been added with **userProgramPC**.

services/approved-studies.js
```diff
@@ -329,10 +337,26 @@ class ApprovedStudiesService {
...
+
         return {...updateStudy, programs: programs, primaryContact: primaryContact};  
     }
+
+    #getConcierge(programs, primaryContact, isProgramPrimaryContact) {
+        // primary contact from the study
+        const [conciergeName, conciergeEmail] = (primaryContact)? [`${primaryContact?.firstName || ""} ${primaryContact?.lastName || ''}`, primaryContact?.email || ""] :
+            ["",""];
+        // isProgramPrimaryContact determines if the program's primary contact should be used.
+        if (isProgramPrimaryContact && programs?.length > 0) {
```
vshand11, line 349: if isProgramPrimaryContact set conciergeName, conciergeEmail to ""  to fix the issue QA addressed.

vshand11 (changes requested):
Please contact either Ming or Alec for the new logic that set conciergeName, conciergeEmail yo "" if isProgramPrimaryContact === true

services/approved-studies.js
knockknockyoo: This is not the solution to fix the issue. I've confirmed with Ming, if you have any other requirement to be addressed, please contact Ming.

vshand11 (approved):
Since you've got confirmed,  approve.

AustinSMueller (approved):

PR #773, opened 2025-04-03
CRDCDH 2483

vshand11:
1) added a new graphQL API, downloadMetadataFile(batchID: String!, fileName: String):String 
2) created a new function in submission.js, async getMetadataFile(params, context)
3) create a new function in batch-service.js, async getMetadataFile(aBatch, fileName)

### DIFF
--- a/services/batch-service.js
+++ b/services/batch-service.js
@@ -12,7 +12,7 @@ const {MongoPagination} = require("../crdc-datahub-database-drivers/domain/mongo
 const {isTrue} = require("../crdc-datahub-database-drivers/utility/string-utility");
 const fs = require('fs');
 const path = require('path');
-const {zipFilesInDir} = require("../utility/io-util");
+const {makeDir, zipFilesInDir} = require("../utility/io-util");
 
 const LOAD_METADATA = "Load Metadata";
 const OMIT_DCF_PREFIX = 'omit-DCF-prefix';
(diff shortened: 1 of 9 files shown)


services/batch-service.js
```diff
@@ -174,6 +178,67 @@ class BatchService {
...
+     */ 
+    async getMetadataFile(aBatch, fileName) {
+        if(fileName){
+            const file = aBatch?.files?.find(f=>f.fileName === fileName && f.status === FILE.UPLOAD_STATUSES.UPLOADED);
+            if(!file){
+                throw new Error(ERROR.FILE_NOT_EXIST);
+            }
+            return await this.s3Service.createDownloadSignedURL(aBatch?.bucketName, aBatch?.filePrefix, fileName) ;
+        }
+        else{
```
knockknockyoo, line 195: I dont think we need else in this case?

vshand11: If FE pass filename, then return single file download url.  else need return whole batch files as zip file downloading url.   So the else is need.

knockknockyoo: Yes, the logic looks correct. I think we can remove the else block in the code to reduce indentation.

vshand11: ok

AustinSMueller (approved):

PR #774, opened 2025-04-04
Crdcdh 2510

vshand11:
Minor change to remove permission verification and allow all login user to retrieve default PBAC configurations.

### DIFF
--- a/services/configurationService.js
+++ b/services/configurationService.js
@@ -21,7 +21,6 @@ class ConfigurationService {
     async getPBACDefaults(params, context){
         verifySession(context)
             .verifyInitialized()
-            .verifyPermission(USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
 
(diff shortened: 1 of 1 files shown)


services/configurationService.js
```diff
@@ -20,8 +20,7 @@ class ConfigurationService {
      */
     async getPBACDefaults(params, context){
         verifySession(context)
-            .verifyInitialized()
-            .verifyPermission(USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
+            .verifyInitialized();
```
knockknockyoo, line 23: I think we should include all roles, at the very least.

vshand11: Based on the description in the task.  All login user can retrieve.   No need any roles.

knockknockyoo (approved):
Approved

AustinSMueller (approved):

PR #780, opened 2025-04-07
Crdcdh 2491

vshand11:
1) Add a record in configuration as listed below:
_id: "941b6ccc-afc3-4a70-ae6d-370faf13dadd"
type: "CLI_UPLOADER_VERSION"
current_version: "4.0"

2) Defince GraphQL API, retrieveCLIUploaderVersion: String
3) implemented API configurationService.retrieveCLIUploaderVersion

### DIFF
--- a/services/configurationService.js
+++ b/services/configurationService.js
@@ -1,6 +1,7 @@
 const {verifySession} = require("../verifier/user-info-verifier");
 const USER_PERMISSION_CONSTANTS = require("../crdc-datahub-database-drivers/constants/user-permission-constants");
-PBAC_CONFIG_TYPE = "PBAC";
+const PBAC_CONFIG_TYPE = "PBAC";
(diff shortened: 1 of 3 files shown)


services/configurationService.js
```diff
@@ -56,6 +57,19 @@ class ConfigurationService {
...
         };
     }
+
+    async  retrieveCLIUploaderVersion(params, context) {
+        verifySession(context)
+            .verifyInitialized();
+        return await this.getCurrentCLIUploaderVersion();
+    }
+
+    async getCurrentCLIUploaderVersion() {
```
knockknockyoo, line 67: Can we make it private?

vshand11: No,  need to be called from other service.

knockknockyoo (approved):
Approved

AustinSMueller (approved):

PR #792, opened 2025-04-14
CRDCDH-2523

vshand11:
Updated the function,  #ProcessSubmissionNodes(result, IDPropName=null),  to fix the issue in the ticket

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -824,9 +824,6 @@ class Submission {
             for (let node of result.results) {
                 if (!returnVal.IDPropName) returnVal.IDPropName = node.IDPropName;
                 if (node?.parents && node.parents.length > 0) {
-                    // for (let parent of node.parents) {
-                    //     node.props[`${parent.parentType}.${parent.parentIDPropName}`] = parent.parentIDValue;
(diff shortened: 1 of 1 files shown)


services/submission.js
```diff
@@ -823,10 +823,17 @@ class Submission {
             
             for (let node of result.results) {
                 if (!returnVal.IDPropName) returnVal.IDPropName = node.IDPropName;
-                if (node.parents && node.parents.length > 0) {
-                    for (let parent of node.parents) {
-                        node.props[`${parent.parentType}.${parent.parentIDPropName}`] = parent.parentIDValue;
-                    }
+                if (node?.parents && node.parents.length > 0) {
+                    // for (let parent of node.parents) {
+                    //     node.props[`${parent.parentType}.${parent.parentIDPropName}`] = parent.parentIDValue;
```
knockknockyoo, line 828: I am not sure why we commented this instead of removing it permanently.

vshand11: cleaned up, thanks!

AustinSMueller (approved):

PR #805, opened 2025-04-18
Authorization Service

AustinSMueller:
Added a central authorization service to check scopes Added scopes to the Federal Lead entries in PBACDefaults.json as examples

### DIFF
--- a/constants/permission-scope-constants.js
+++ b/constants/permission-scope-constants.js
@@ -3,6 +3,6 @@ module.exports = Object.freeze({
     NONE: "none",
     OWN: "own",
     STUDY: "study",
-    DC: "dc",
+    DC: "DC",
     ROLE: "role"
 });
--- a/constants/permission-scope-constants.js
+++ b/constants/permission-scope-constants.js
@@ -1,0 +1,8 @@
+module.exports = Object.freeze({
+    ALL: "all",
+    NONE: "none",
+    OWN: "own",
+    STUDY: "study",
+    DC: "dc",
+    ROLE: "role"
+});
--- a/resources/json/PBACDefaults_config.json
+++ b/resources/json/PBACDefaults_config.json
@@ -7,161 +7,144 @@
         "role": "Federal Lead",
         "permissions": [
           {
-            "_id": "submission_request:view",
+            "_id": "submission_request:view:all",
             "group": "Submission Request Permissions",
             "name": "View",
             "order": 1,
             "checked": true,
             "disabled": false,
-            "scopes": ["all"]
           },
           {
-            "_id": "submission_request:create",
+            "_id": "submission_request:create:all",
             "group": "Submission Request Permissions",
             "name": "Create",
             "inherited": ["submission_request:view"],
             "order": 2,
             "checked": false,
-            "disabled": false,
-            "scopes": ["all"]
+            "disabled": false
           },
           {
-            "_id": "submission_request:review",
+            "_id": "submission_request:review:all",
             "group": "Submission Request Permissions",
             "name": "Review",
             "inherited": ["submission_request:view"],
             "order": 4,
             "checked": false,
             "disabled": false,
-            "scopes": ["all"]
           },
           {
-            "_id": "submission_request:submit",
+            "_id": "submission_request:submit:all",
             "group": "Submission Request Permissions",
             "name": "Submit",
             "order": 3,
             "checked": false,
-            "disabled": false,
-            "scopes": ["all"]
+            "disabled": false
           },
           {
-            "_id": "submission_request:cancel",
+            "_id": "submission_request:cancel:all",
             "group": "Submission Request Permissions",
             "name": "Cancel/Restore",
             "order": 5,
             "checked": true,
-            "disabled": false,
-            "scopes": ["all"]
+            "disabled": false
           },
           {
-            "_id": "access:request",
+            "_id": "access:request:none",
             "group": "Miscellaneous",
             "name": "Request Access",
             "order": 1,
             "checked": false,
-            "disabled": true,
-            "scopes": ["none"]
+            "disabled": true
           },
           {
-            "_id": "data_submission:view",
+            "_id": "data_submission:view:study",
             "group": "Data Submission Permissions",
             "name": "View",
             "order": 1,
             "checked": true,
-            "disabled": false,
-            "scopes": ["study"]
+            "disabled": false
           },
           {
-            "_id": "data_submission:create",
+            "_id": "data_submission:create:study",
             "group": "Data Submission Permissions",
             "inherited": ["data_submission:view"],
             "name": "Create",
             "order": 2,
             "checked": false,
-            "disabled": false,
-            "scopes": ["study"]
+            "disabled": false
           },
           {
-            "_id": "data_submission:cancel",
+            "_id": "data_submission:cancel:study",
             "group": "Data Submission Permissions",
             "name": "Cancel",
             "order": 3,
             "checked": false,
-            "disabled": false,
-            "scopes": ["study"]
+            "disabled": false
           },
           {
-            "_id": "data_submission:review",
+            "_id": "data_submission:review:study",
             "group": "Data Submission Permissions",
             "name": "Review",
             "inherited": ["data_submission:view"],
             "order": 4,
             "checked": false,
-            "disabled": false,
-            "scopes": ["study"]
+            "disabled": false
           },
           {
-            "_id": "data_submission:admin_submit",
+            "_id": "data_submission:admin_submit:study",
             "group": "Data Submission Permissions",
             "name": "Admin Submit",
             "order": 5,
             "checked": false,
-            "disabled": true,
-            "scopes": ["study"]
+            "disabled": true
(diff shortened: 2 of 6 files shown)


test/services/authorization.service.test.js
```diff
@@ -0,0 +1,81 @@
...
+            permissions: [
+                "data_submission:cancelled:none",
+                'data_submission:submitted:all'
+            ]
+        };
+    });
+
+    test("/test read scope from permission", () => {
+        let permission = "data_submission:submitted"
+        expect(authorizationService.getPermissionScope(user, permission)).toStrictEqual(["all"])
```
knockknockyoo, line 44: I am trying to understand. When you insert multiple permissions, why is returning only **all**? There are different permissions.

services/authorization-service.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Function to get a user's scopes for a given permission
+     * @param user - the current user
+     * @param permission - the name of the permission without scopes, please use a permissions constant for this input
+     * Example: "entity:action"
+     * @returns {*|string[]|*[]} - an array containing the scopes for the permission or an empty array, all scopes will be lowercase
+     */
+    getPermissionScope(user, permission){
```
n2iw, line 16: We also need to return scope value(s). For example, if scope is **study**, then we need to return a list of study ids associated with the user.

n2iw (changes requested):

services/authorization-service.js
```diff
@@ -0,0 +1,86 @@
...
+ * example: "entity:action:scope1+scope2"
+ * @returns {{permission: string, scopes: string[]}|null} - an object containing the permission string without scopes and
+ * an array containing the scopes
+ */
+function parsePermissionString(permissionString){
+    if (!permissionString || typeof permissionString !== "string"){
+        throw new Error(ERROR.INVALID_PERMISSION_STRING);
+    }
+    const permissionStringElements = permissionString.split(':');
+    if (!permissionStringElements || permissionStringElements?.length < 2) {
```
knockknockyoo, line 69: If you just add ":::", it will return empty one. Should we prevent that?

services/authorization-service.js
```diff
@@ -0,0 +1,86 @@
...
+        return null;
+    }
+    let permission = [permissionStringElements[0],permissionStringElements[1]].join(":");
+    let scopes = [];
+    if (permissionStringElements.length > 2){
+        scopes = permissionStringElements[2].split('+');
```
knockknockyoo, line 75: If the value "++++", it will return empty value. Should we prevent that?

test/services/authorization.service.test.js
AustinSMueller: This is testing that it reads from the mock response. The mock response is a stand-in for reading from the database so this test is really just making sure that when no scopes are provided that the output matches the specified mock response for the database call

services/authorization-service.js
AustinSMueller: I'll do some testing on this but I think it's ok. The result would be an empty array which is the same response as when the user does not have the permission.

services/authorization-service.js
AustinSMueller: I think this is ok as well, even if we were to check for this the returned result would still be an empty array

services/authorization-service.js
AustinSMueller: I'll work on this change. I might have some questions but I will reach out if I need any help

services/authorization-service.js
knockknockyoo: Yes. The permission input wont interrupt the output. Thanks.

services/authorization-service.js
knockknockyoo: Yes. The permission input wont interrupt the output. Thanks.

services/authorization-service.js
```diff
@@ -0,0 +1,86 @@
...
+     * @returns {*|string[]|*[]} - an array containing the scopes for the permission or an empty array, all scopes will be lowercase
+     */
+    getPermissionScope(user, permission){
+        const userPermissions = user?.permissions;
+        if (!userPermissions || !permission){
+            return [];
+        }
+        // Loop through the user's permissions until one matching the input permission is found
+        for (const userPermission of userPermissions){
+            let permissionAndScope = parsePermissionString(userPermission);
```
knockknockyoo, line 23: I think permissionAndScope should be constant?

knockknockyoo:
Thank you Austin. This is only change needed"**await** this.configurationService.getPBACByRoles"

services/authorization-service.js
AustinSMueller: This has been updated to be a constant

services/authorization-service.js
```diff
@@ -0,0 +1,86 @@
...
+                    return scopes;
+                }
+                /*
+                The below block of code is for backwards compatibility. I expect this will eventually be removed.
+                If a permission is found but no scopes are specified then this will retrieve and use the default
+                scopes from the PBAC configuration in MongoDB.
+                */
+                scopes = [];
+                const userRole  = user?.role;
+                let pbacDefaults = this.configurationService.getPBACByRoles([userRole]);
```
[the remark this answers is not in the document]
AustinSMueller: good catch, I've added the await. I've also made the function asynchronous and updated the test calls as well

services/authorization-service.js
AustinSMueller: I've rewrote the getPermissionScope function to return {scopes: string[], scopeValues: string[]}

constants/permission-scope-constants.js
```diff
@@ -0,0 +1,8 @@
+module.exports = Object.freeze({
+    ALL: "all",
+    NONE: "none",
+    OWN: "own",
+    STUDY: "study",
+    DC: "dc",
```
n2iw, line 6: I think the value should also be "DC", since it's an acronym. Peter is using "DC" in the PBAC JSON file also.

services/authorization-service.js
```diff
@@ -0,0 +1,104 @@
...
+        // Loop through the user's permissions until one matching the input permission is found
+        for (const userPermission of userPermissions){
+            const permissionAndScope = parsePermissionString(userPermission);
+            let scopes = permissionAndScope?.scopes;
+
+            if (permissionAndScope?.permission === permission) {
+                if (!!scopes && scopes.length > 0){
+                    output.scopes = scopes;
+                    output.scopeValues = permissionAndScope?.scopeValues || []
+                    return output;
```
n2iw, line 37: I suggest the return to be restructured to something like below, so when there are multiple scopes and each has different scope values, the return is easy to read.
`[
    {
        scope: "study",
        scopeValues: [
            "study_001",
            "study_002"
        ]
    },
    {
        scope: "DC",
        scopeValues: [
            "ICDC",
            "CTDC"
        ]
    }
]`

services/authorization-service.js
```diff
@@ -0,0 +1,104 @@
...
+        }
+        // Loop through the user's permissions until one matching the input permission is found
+        for (const userPermission of userPermissions){
+            const permissionAndScope = parsePermissionString(userPermission);
+            let scopes = permissionAndScope?.scopes;
+
+            if (permissionAndScope?.permission === permission) {
+                if (!!scopes && scopes.length > 0){
+                    output.scopes = scopes;
+                    output.scopeValues = permissionAndScope?.scopeValues || []
```
n2iw, line 36: When scope is "study", the scopeValues should be read from user's assigned studies. Similarly, when scope is "DC", scopeValues should be read from user's assigned data commons.

n2iw (changes requested):

services/authorization-service.js
```diff
@@ -7,59 +7,70 @@ class AuthorizationService {
...
+     * for this input
      * Example: "entity:action"
-     * @returns {*|string[]|*[]} - an array containing the scopes for the permission or an empty array, all scopes will be lowercase
+     * @returns {{scopes: string[], scopeValues: string[]}} an object containing the user's scopes and scope values for
+     * the input permission. If the user does not have the permission or has no scopes for the permission, then the
+     * default response is {scopes: ["none"], scopeValues: []}
      */
     async getPermissionScope(user, permission){
+        let output = {
+            scopes: ['none'],
```
knockknockyoo, line 21: **none** should be read from the constant you created.

knockknockyoo (approved):
Approved!

resources/json/PBACDefaults_config.json
```diff
@@ -71,15 +78,17 @@
...
           },
           {
             "_id": "data_submission:cancel",
             "group": "Data Submission Permissions",
             "name": "Cancel",
             "order": 3,
             "checked": false,
-            "disabled": false
+            "disabled": false,
+            "scopes": ["study"]
```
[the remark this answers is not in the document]
AustinSMueller: I think you're right, I've pushed an update to move the default scopes and scopeValues to be included in the _id string. The code has been updated to read from defaults from there as well

vshand11 (approved):
LGTM

n2iw (approved):
LGTM

PR #816, opened 2025-04-25
CRDCDH-2623

knockknockyoo:


AustinSMueller (changes requested):

services/user.js
```diff
@@ -429,6 +449,22 @@ class UserService {
         return { ...prevUser, ...userAfterUpdate};
     }
 
+    async #isFederalLeadPermission(userInfo) {
+        const validScopes = await this.authorizationService.getPermissionScope(userInfo, USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
+        return UserScope.create(validScopes)
```
knockknockyoo, line 454: I created the UserScope class to verify the code. Isn't that sufficient?

constants/error-constants.js
```diff
@@ -173,6 +173,7 @@ const ERROR = {
     INVALID_APPROVED_STUDIES_ACCESS_REQUEST: "Failed to request an access request because of invalid or missing approved study IDs.",
     DUPLICATE_ORGANIZATION_NAME: "Duplicate organization name found: $item$",
     NO_ADMIN_USER: "No admin user found or no PBAC request access notification is enabled.",
+    INVALID_FEDERAL_LEAD_REQUEST: "You are have limited access to the Federal Lead users",
```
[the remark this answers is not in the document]
knockknockyoo: I have updated the error message.

services/user.js
```diff
@@ -349,6 +363,12 @@ class UserService {
         if (!user || !Array.isArray(user) || user.length < 1 || user[0]?._id !== params.userID) {
             throw new Error(SUBMODULE_ERROR.USER_NOT_FOUND);
         }
+
+        const isFederalLeadOnly = await this.#isFederalLeadPermission(context?.userInfo);
```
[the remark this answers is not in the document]
knockknockyoo: @n2iw  I have renamed the function. @AustinSMueller  Please, confirm this is the only case ROLES.FEDERAL_LEAD explicitly used in the scope.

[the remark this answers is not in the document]
knockknockyoo: @n2iw I have removed the Federal Lead role.

services/user.js
```diff
@@ -200,14 +201,22 @@ class UserService {
...
-            .verifyInitialized()
-            .verifyPermission(USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
-        
+            .verifyInitialized();
+        await this.#isPermitted(context?.userInfo, USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
+
         if (!params?.userID) {
             throw new Error(SUBMODULE_ERROR.INVALID_USERID);
         }
+        const isFederalLeadOnly = await this.#isManageUserScope(context?.userInfo, ROLES.FEDERAL_LEAD);
```
n2iw, line 210: We should not check if the allowed role is Federal Lead at all, because it might be other roles in the future. Instead, we just need to know if the scope is "role" or "all", then use this information to create filters.

services/user.js
```diff
@@ -200,14 +201,22 @@ class UserService {
...
+        if (context?.userInfo?.role === ROLES.FEDERAL_LEAD && !isFederalLeadOnly) {
```
n2iw, line 211: Again, we should not check Federal Lead role.

services/user.js
```diff
@@ -200,14 +201,22 @@ class UserService {
...
+            throw new Error(ERROR.INVALID_FEDERAL_LEAD_REQUEST);
+        }
+
+        const filters = {
+            _id: params.userID,
+            ...(isFederalLeadOnly ? { role: ROLES.FEDERAL_LEAD } : {})
```
n2iw, line 217: Here, if the scope we got is "role" then use the scope value in the filters. If the scope is "all" then no filter is needed.

services/user.js
```diff
@@ -227,13 +236,19 @@ class UserService {
...
-            .verifyInitialized()
-            .verifyPermission(USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
+            .verifyInitialized();
+        await this.#isPermitted(context?.userInfo, USER_PERMISSION_CONSTANTS.ADMIN.MANAGE_USER);
+        const isFederalLeadOnly = await this.#isManageUserScope(context?.userInfo, ROLES.FEDERAL_LEAD);
 
-        const result = await this.userCollection.aggregate([{
-            "$match": {}
-        },]);
+        if (context?.userInfo?.role === ROLES.FEDERAL_LEAD && !isFederalLeadOnly) {
```
n2iw, line 243: Same as above, we should not check Federal Lead role at all.

services/user.js
```diff
@@ -227,13 +236,19 @@ class UserService {
...
+            return [];
+        }
 
+        const result = await this.userCollection.aggregate([{
+            "$match": {
+                ...(isFederalLeadOnly ? { role: ROLES.FEDERAL_LEAD } : {})
```
n2iw, line 249: Same as above, if scope is "all" then no filters. If scope is "role", then use the scope value returned by authorization function to create the filter.

services/user.js
```diff
@@ -349,6 +363,12 @@ class UserService {
         if (!user || !Array.isArray(user) || user.length < 1 || user[0]?._id !== params.userID) {
             throw new Error(SUBMODULE_ERROR.USER_NOT_FOUND);
         }
+
+        const isFederalLeadOnly = await this.#isManageUserScope(context?.userInfo, ROLES.FEDERAL_LEAD);
```
n2iw, line 367: Same as above

n2iw (changes requested):

services/user.js
knockknockyoo: Yes, the specific role is not removed.

services/user.js
knockknockyoo: The Federal role is removed.

services/user.js
knockknockyoo: Resolved.

services/user.js
knockknockyoo: Yes, the filter is applied for the role scope.

n2iw (approved):
LGTM

PR #821, opened 2025-05-01
CRDCDH-2659

knockknockyoo:


services/application.js
```diff
@@ -144,7 +150,10 @@ class Application {
     async saveApplication(params, context) {
         verifySession(context)
             .verifyInitialized()
-            .verifyPermission(USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.CREATE);
+        const userScope = await this.#getUserScope(context?.userInfo, USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.CREATE);
```
n2iw, line 153: This should be just checking if the user is creator of the application, not create permission.

services/application.js
```diff
@@ -197,29 +210,21 @@ class Application {
 
         const baseConditions = {...statusCondition, ...programNameCondition, ...studyNameCondition, ...submitterNameCondition};
         return (() => {
-            switch (userRole) {
-                case ROLES.ADMIN:
-                case ROLES.FEDERAL_LEAD:
-                case ROLES.DATA_COMMONS_PERSONNEL:
+            switch (true) {
```
n2iw, line 213: Using switch on a true value is weird, please just use if-else to test this.

services/application.js
```diff
@@ -318,8 +334,12 @@ class Application {
 
     async reopenApplication(document, context) {
         verifySession(context)
-            .verifyInitialized()
-            .verifyPermission(USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.CREATE);
+            .verifyInitialized();
+        const userScope = await this.#getUserScope(context?.userInfo, USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.CREATE);
```
n2iw, line 338: This should be just check if user is creator of the application, not create permission.

services/application.js
```diff
@@ -389,14 +418,14 @@ class Application {
             throw new Error(ERROR.INVALID_APPLICATION_RESTORE_STATE);
         }
         const userInfo = context?.userInfo;
-        const isEnabledPBAC = userInfo?.permissions?.includes(USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.CANCEL);
-        const isPowerRole = [ROLES.FEDERAL_LEAD, ROLES.ADMIN, ROLES.DATA_COMMONS_PERSONNEL].includes(userInfo?.role);
-
-        const isNonPowerRole = [ROLES.USER, ROLES.SUBMITTER].includes(userInfo?.role);
+        const userScope = await this.#getUserScope(context?.userInfo, USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.REVIEW);
```
n2iw, line 421: Isn't restore and cancel using the same permission? Why change to check review permission?

n2iw (changes requested):

services/application.js
knockknockyoo: Changed both data submission/submission request.

services/application.js
knockknockyoo: My apology. This should be the Cancel permission.

services/application.js
knockknockyoo: Yes, updated the submission request creator only.

services/application.js
knockknockyoo: For saveApplication API, it will be two scenarios. The codes is updated.
1. store the submission request if it does not exist; I think it should be the **Create** permission.
2. if the submission request existed, it checks only the creator.

n2iw (approved):
LGTM

PR #845, opened 2025-05-19
CRDCDH-2710

knockknockyoo:


AustinSMueller (approved):

services/submission.js
```diff
@@ -1469,20 +1469,43 @@ class Submission {
             return aSubmission;
         }
 
+        return await this.#resetValidation(aSubmission, version);
+    }
+
+    async #resetValidation(aSubmission, version){
```
n2iw, line 1475: The update version portal is nothing related to reset validation (except it's one of the potential reason why we want to reset validation), so it should remain in updateSubmissionModelVersion function. Then resetValidation is reusable in other situations.

n2iw (changes requested):

services/submission.js
knockknockyoo: I've separated the logic to update the data model. Please, review it again.

n2iw (approved):
LGTM

PR #847, opened 2025-05-19
PBAC Refactoring BUGs

knockknockyoo:
1. accept all scopes values
2. delete application / restore application only for the user.
3. data submission
- own scope needs to be the user-session.

### DIFF
--- a/services/application.js
+++ b/services/application.js
@@ -424,7 +424,7 @@ class Application {
         }
         // User owned application
         const isApplicationOwned = userInfo?._id === aApplication?.applicant?.applicantID;
-        if (userScope.isNoneScope() || (userScope.isOwnScope() && !isApplicationOwned)) {
+        if (userScope.isOwnScope() && !isApplicationOwned) {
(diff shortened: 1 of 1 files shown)


services/application.js
```diff
@@ -418,13 +419,12 @@ class Application {
         }
         const userInfo = context?.userInfo;
         const userScope = await this.#getUserScope(context?.userInfo, USER_PERMISSION_CONSTANTS.SUBMISSION_REQUEST.CANCEL);
-        if (userScope.isNoneScope() || !(userScope.isOwnScope() || userScope.isAllScope())) {
+        if (userScope.isNoneScope() || !(userScope.isOwnScope() && userScope.isAllScope())) {
```
n2iw, line 422: Following part doesn't look right to me. I assume user will never have own scope and all scope at the same time, so inside the parenthesis is always false?
`!(userScope.isOwnScope() && userScope.isAllScope())`

n2iw (approved):
LGTM

PR #870, opened 2025-05-28
CRDCDH-2835

knockknockyoo:


n2iw (changes requested):
1.  Line 970 "data_submission:view:own" needs to be changed to "data_submission:view:study".
2. Will FE and BE code both handle "inherited": []? maybe consider just delete the line entirely?

n2iw (approved):
LGTM

PR #898, opened 2025-06-16
CRDCDH-2810

knockknockyoo:


AustinSMueller:
Based on the conversation this morning let's remove the private function prefix for the names. Also can you please add some unit testing, with a focus on invalid or null response from the aggregation pipeline requests? With pipelines that complicated we need to make sure that we can handle errors if something goes wrong. Unit tests are not needed for the private functions unless you want to write/generate them

AustinSMueller (changes requested):
please see my other comments for the requested changes

PR #900, opened 2025-06-17
Crdcdh 2947

vshand11:
Update private methods/variables under services classes.  Tested in local with postman.


knockknockyoo:
I understood from the chat that we should keep the private method using an underscore (_) instead of a hash (#).

PR #938, opened 2025-07-10
FixGetDiagnosisAge

vshand11:
Minor change.  Force to use participant.nodeID for get diagnosis age that fixes the get age in case the dbGap_participant_ID is not null.


n2iw (approved):
LGTM

PR #956, opened 2025-07-16
CRDCDH 3001

vshand11:
Implemented CRDCDH 3001.

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -717,20 +717,18 @@ class Submission {
         if (isNotPermitted) {
             throw new Error(ERROR.INVALID_VALIDATE_METADATA)
         }
-        // start validation, change validating status
-        const [prevMetadataValidationStatus, prevFileValidationStatus, prevCrossSubmissionStatus, prevTime] =
-            [aSubmission?.metadataValidationStatus, aSubmission?.fileValidationStatus, aSubmission?.crossSubmissionStatus, aSubmission?.updatedAt];
-
         // if the user role is DATA_COMMONS_PERSONNEL, and the submission status is "Submitted", and aSubmission?.crossSubmissionStatus is "Failed", 
         // and params.type contains validation type metadata, replace validation type metadata to cross-submission validation.  User story CRDCDH-2830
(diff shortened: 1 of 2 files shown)


services/submission.js
```diff
@@ -717,10 +717,18 @@ class Submission {
         if (isNotPermitted) {
             throw new Error(ERROR.INVALID_VALIDATE_METADATA)
         }
+        // if the user role is DATA_COMMONS_PERSONNEL, and the submission status is "Submitted", and aSubmission?.crossSubmissionStatus is "Error", 
+        // and params.types not contains CROSS_SUBMISSION, add CROSS_SUBMISSION. User story CRDCDH-2830
+        if (userInfo?.role === ROLES.DATA_COMMONS_PERSONNEL && aSubmission?.status === SUBMITTED &&
```
n2iw, line 722: If it's needed to check permission, please check permission instead of role. I assume cross-validation requires data_submission:review permission?

services/submission.js
```diff
@@ -717,10 +717,18 @@ class Submission {
         if (isNotPermitted) {
             throw new Error(ERROR.INVALID_VALIDATE_METADATA)
         }
+        // if the user role is DATA_COMMONS_PERSONNEL, and the submission status is "Submitted", and aSubmission?.crossSubmissionStatus is "Error", 
+        // and params.types not contains CROSS_SUBMISSION, add CROSS_SUBMISSION. User story CRDCDH-2830
+        if (userInfo?.role === ROLES.DATA_COMMONS_PERSONNEL && aSubmission?.status === SUBMITTED && 
+            aSubmission?.crossSubmissionStatus === VALIDATION_STATUS.ERROR && params?.types &&
+            !params?.types?.includes(VALIDATION.TYPES.CROSS_SUBMISSION)) {
```
n2iw, line 724: Do we need to exclude CROSS_SUBMISSION here, since it triggers cross-validation anyways?

n2iw (changes requested):

services/submission.js
vshand11: updated.

PR #1032, opened 2025-08-19
CRDCDH 3054

vshand11:
Use data fields as query parameters.


n2iw (approved):
LGTM, will merge and test.

PR #1048, opened 2025-08-22
CRDCDH-3056

wfy1997:
new API updateSubmissionInfo to edit submission
The API updateSubmissionInfo can update modelVersion, name, submitterID

### DIFF
--- a/services/submission.js
+++ b/services/submission.js
@@ -1623,11 +1623,11 @@ class Submission {
      * @param {*} context
      * @returns {Promise<Submission>}
      */
-    async updateSubmissionModelVersion(params, context) {
+    async updateSubmissionInfo(params, context) {
         verifySession(context)
             .verifyInitialized();
 
-        const {_id, version, submitterID} = params;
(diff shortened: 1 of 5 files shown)


services/submission.js
```diff
@@ -1618,83 +1618,108 @@ class Submission {
...
-                if (submitterID) {
-                    const newSubmitter = await this.userDAO.findFirst({id: submitterID});
-                    const preSubmitter = await this.userDAO.findFirst({id: aSubmission?.submitterID});
-                    return {prevSubmitter: preSubmitter, newSubmitter: newSubmitter};
-                }
-                return {};
-            })()
-        ]);
-
+        const userScope = await this._getUserScope(context?.userInfo, USER_PERMISSION_CONSTANTS.DATA_SUBMISSION.REVIEW, aSubmission);
```
n2iw, line 1636: Next line (1637) will block submitters from using this API, because they won't have review permission, so userScope will be None.

n2iw (changes requested):
I think it's probably better to separate change submission name to another API. That way it's much easier to do permission control.

PR #1067, opened 2025-09-02
CRDCDH-3056

wfy1997:
Reduced the character limit for the submission name to 25.
Made Admin, DCP, or Federal Lead able to update the model version. 
Added submitterName into the return object.
Bypassed the reset validation if model version not changed.


amattu2 (changes requested):
Discussed in Slack, but commenting here for visibility:

The updateSubmissionInfo API is only allowing updates in New/In Progress, the requirements allow for updates in other statuses too. This issue isn't related to this particular set of changes, but let's address that in this PR too. Thanks!

amattu2 (approved):

PR #1073, opened 2025-09-04
CRDCDH-3056

wfy1997:
Update the testing to remove expect _resetValidation since only submitter ID changing will no longer trigger reset


n2iw (approved):
LGTM

PR #1093, opened 2025-09-23
modified build be workflow

tracy18:
modified build be workflow


n2iw (approved):
LGTM, let's merge and test.

PR #1100, opened 2025-10-01
Make GPAName and dbGaPID optional parameters

AustinSMueller:
Make the GPAName and dbGaPID optional parameters createApprovedStudy and updateApprovedStudy. Return an error if updating or creating a study with an invalid GPAName state. If GPAName is null or empty and controlledAccess is true then isPendingGPA must be true. If this validation fails then an error is returned.


n2iw (approved):
LGTM

PR #1156, opened 2026-01-28
3481: Only store UserID in API token

AustinSMueller:
- updated the token generation code to only store the user's ID
- backwards compatible to accept existing tokens
- updated unit tests accordingly

### DIFF
--- a/services/user.js
+++ b/services/user.js
@@ -34,9 +34,11 @@ const isValidUserStatus = (userStatus) => {
     if (userStatus && !validUserStatus.includes(userStatus)) throw new Error(SUBMODULE_ERROR.INVALID_USER_STATUS);
 }
 
-const createToken = (userInfo, token_secret, token_timeout)=> {
+const createToken = (userID, token_secret, token_timeout)=> {
     return jwt.sign(
-        userInfo,
+        // sub (Subject) is used to follow JWT naming conventions
+        // https://www.iana.org/go/rfc7519#section-4.1.2
+        { sub: userID },
         token_secret,
         { expiresIn: token_timeout });
 }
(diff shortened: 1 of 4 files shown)


services/user.js
```diff
@@ -133,10 +135,14 @@ class UserService {
     async grantToken(params, context){
         isLoggedInOrThrow(context);
         isValidUserStatus(context?.userInfo?.userStatus);
+        if (!context?.userInfo?._id) {
```
n2iw, line 138: Can we use .id instead of ._id here?

services/user.js
```diff
@@ -133,10 +135,14 @@ class UserService {
...
+            console.error("Cannot create a token because the User ID is missing from the context");
+            throw new Error(SUBMODULE_ERROR.INVALID_USERID);
+        }
         if(context?.userInfo?.tokens){
             context.userInfo.tokens = []
         }
-        const accessToken = createToken(context?.userInfo, config.token_secret, config.token_timeout);
+        const accessToken = createToken(context.userInfo._id, config.token_secret, config.token_timeout);
```
n2iw, line 145: Similarly, can we use .id instead of ._id here?

AustinSMueller: We can update this but this will be a large change. The _id field is initialized using the _id field from the users collection in MongoDB. I ran a search and to rename this field we would have to update 74 instances where this field is accessed across the project.

If this change would just be for naming convention purposes I would recommend that we leave it as is. There are no technical or implementation obstacles preventing us from this change, I am just concerned about the scope of the change.

n2iw (approved):
LGTM

PR #1173, opened 2026-03-31
3637: set image de-id flag when creating study

AustinSMueller:
- added pendingImageDeIdentification property to ApprovedStudy
- added pendingImageDeIdentification to createApprovedStudy and approveApplication mutations


n2iw (approved):
LGTM

PR #1185, opened 2026-04-14
Update pending image identification text

AustinSMueller:
- The API error for pending image de-identification is used exactly as returned for display on the frontend. Updated to the display text from the requirements


amattu2 (approved):
LGTM

PR #1198, opened 2026-04-30
3737: Updates after testing

AustinSMueller:
- remove missing study abbreviation mapping from listPrograms (not necessary)
- removed applyStudyAbbreviationFallbackToListPrograms
- revert change that removed using study name as the default study abbreviation when it was not provided


n2iw (approved):
LGTM

PR #1202, opened 2026-05-06
3647: escape literals in search

AustinSMueller:
- created a function to escape regex operators in search inputs so that they are treated as literals. These regex operators where causing parsing errors in the MongoDB queries
- used this escapeRegexLiteral in all queries where the user provides text as a search input
- updated unit tests accordingly


n2iw (approved):
LGTM

